PR #4, opened 2017-05-16
BootstrapTree update to handle tag tree

rlbabyuk:
Made changes BootstrapTree to handle tag tree, this is needed for new tests and group conversion to widgetastic.


mfalesni:
@rlbabyuk DId you check it against trees in CFME? Does it still work with normal trees?

rlbabyuk:
@mfalesni As I found, we are using this tree also in ManageIQTree, so I've tested it with my changes for catalog_items tests. Tests work fine.

mfalesni:
@rlbabyuk  Good. I will need to build a test page(s) for the widgets so we can have some sort of unit tests around that :) I will merge this, please keep an eye on the trees or complints about trees in following days.

rlbabyuk:
@mfalesni sure, thanks:)

PR #5, opened 2017-05-24
Added SelectorDropdown

mfalesni:


mfalesni:
This seems to work though I can't manage to get a working unit test in there. We will see how it does when used on real page.

PR #6, opened 2017-06-12
[RFR] Add about modal widget and test page

mshriver:
In progress about modal and a test page for it

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -11,7 +11,7 @@ from collections import namedtuple
 from widgetastic.exceptions import NoSuchElementException, UnexpectedAlertPresentException
 from widgetastic.log import call_sig
 from widgetastic.utils import ParametrizedLocator, VersionPick
-from widgetastic.widget import ClickableMixin, TextInput, Widget, View, Checkbox, \
+from widgetastic.widget import ClickableMixin, TextInput, Text, Widget, View, Checkbox, \
     do_not_read_this_widget
 from widgetastic.xpath import quote
 
@@ -1382,7 +1382,7 @@ class BootstrapSwitch(Checkbox):
         return self.browser.is_selected(parent=self, locator=self.input)
 
 
-class AboutModal(View):
+class AboutModal(Widget):
     """
     Represents the patternfly about modal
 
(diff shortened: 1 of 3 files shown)


testing/testing_page.html
```diff
@@ -34,5 +34,35 @@
...
+          <h1>Widgetastic About Modal</h1>
+          <div class="product-versions-pf">
+            <ul class="list-unstyled">
+              <li><strong>Field1</strong> Field1Value</li>
+              <li><strong>Field2</strong> Field2Value</li>
+              <li><strong>Field3</strong> Field3Value</li>
+            </ul>
+          </div>
+          <div class="trademark-pf">
+            Widgetsatic Trademark and Copyright Information
```
mfalesni: typo.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1380,3 +1380,58 @@ def __init__(self, parent, id=None, name=None, logger=None):
...
+class AboutModal(Widget):
+    """
+    Represents the patternfly about modal
+
+    Provides a close method
+    """
+    ROOT = ParametrizedLocator('//div[normalize-space(@id)={@id|quote} and '
+                               'contains(@class, "modal") and contains(@class, "fade")]')
+    CLOSE_LOC = './/div[@class="modal-header"]/button[@class="close" and @data-dismiss="modal"]'
+    TITLE_LOC = './/div[@class="modal-body"]/h2'
```
mfalesni: I believe if you define this as a Text widget, it should work as well. Similar with some of the other locators.

testing/test_about_modal.py
```diff
@@ -0,0 +1,42 @@
...
+        button = Button('Launch about modal')
+        about = AboutModal(id=modal_id)
+
+    view = TestView(browser)
+    assert not view.about.is_open
+
+    # Open the modal
+    assert view.button.title == 'Launch Modal'
+    assert not view.button.disabled
+    view.button.click()
```
mshriver: This doesn't seem to be working in PhantomJS so travis won't pass.

src/widgetastic_patternfly/__init__.py
mshriver: Asked about this offline as well.

I'm not following the implementation you're describing here.  Tried defining a `self.title_widget` within the constructor using a Text widget, but then just have a widget_descriptor when I try to access it in the property.

Not seeing any other locators that are implemented as Text widgets?

mfalesni: Widgets can only be defined on the class, not in constructor.

PR #10, opened 2017-07-03
Add a widget for the Bootstrap nav component

rsnyman:
This new widget encapsulates a Bootstrap nav component and provides methods for selecting an item and checking if an item is disabled.

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -393,6 +393,79 @@ class NavDropdown(Widget, ClickableMixin):
         return '{}({!r})'.format(type(self).__name__, self.locator)
 
 
+class BootstrapNav(Widget):
+    """Encapsulate a Bootstrap nav component
+
+    PatternFly is based on Bootstrap, and thus many of the Bootstrap components are available to
+    PatternFly users. This widget provides convenience methods for the Bootstrap nav component for
+    clicking on links and determining if an item in the nav is disabled.
+
+    When instantiating this widget, use the XPath locator to point to exactly which Bootstrap nav
+    you wish to work with.
+
+    .. _code:: python
+
+       nav = BootstrapNav('//div[id="main"]/ul[@contains(@class, "nav")]')
+
+    See http://getbootstrap.com/components/#nav for more information on Bootstrap nav components.
+    """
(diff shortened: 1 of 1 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -393,6 +393,68 @@ def __repr__(self):
...
+        Widget.__init__(self, parent, logger=logger)
+        self.locator = locator
+
+    def __repr__(self):
+        """String representation of this object"""
+        return '{}({!r})'.format(type(self).__name__, self.locator)
+
+    def __locator__(self):
+        """Return the locator for this widget"""
+        return self.locator
```
mfalesni: Replace this method with:
```python
ROOT = ParametrizedLocator('{@locator}')
```

on the class.

src/widgetastic_patternfly/__init__.py
```diff
@@ -393,6 +393,68 @@ def __repr__(self):
         return '{}({!r})'.format(type(self).__name__, self.locator)
 
 
+class BootstrapNav(Widget):
+    """Encapsulate a Bootstrap nav component"""
```
mfalesni: Maybe a couple more words about it, also where locator should point exactly, maybe a link to the component itself ...

src/widgetastic_patternfly/__init__.py
```diff
@@ -393,6 +393,68 @@ def __repr__(self):
...
+    CURRENTLY_SELECTED = './/li[contains(@class, "active")]/a'
+    TEXT_MATCHING = './/li/a[text()={txt}]'
+    ATTR_MATCHING = './/li/a[@{attr}={txt}]'
+    TEXT_DISABLED = './/li[contains(@class, "disabled")]/a[text()={txt}]'
```
mfalesni: text query should be encapsulated in `normalize-space(...)`. I would personally prefer using `.` instead of `text()` as `.` match the text of whole substructure of that element, not just text of that element. Unless there is some other text which would make it difficult :)

rsnyman: There's nothing else inside the `a` element, so `normalize-space(...)` doesn't achieve anything more than `text()`.

rsnyman:
@mfalesni Is there anything more you'd like me to do with this PR in order to get it merged?

PR #11, opened 2017-07-19
[RFR] Tab's ROOT doesn't point out to its contents

izapolsk:
Widgets under Tab cannot be found due to that issue. 
(This issue wasn't encountered before because parent of parent wasn't passed correctly)


izapolsk:
there is one failure related to modal. it isn't related to my change

mfalesni:
INDIRECT should ensure that the Tab is not used as the root for lookup since tabs can be completely separate away from widgets.

PR #13, opened 2017-08-07
[RFR] BootstrapSwitch incorrect gets switch state in some cases

izapolsk:
It seems this is patternfly or manageiq bug. 
For example there is BootstrapSwitch on Middleware->Servers->Add Deployment page, it is in "on" state by default but selenium's is_selected method will return False until it gets clicked.
The issue can be noticed in JS as well.
el = document.getElementById('force_deployment_cb')
el.checked (returns False though it is True)

This PR is a workaround for above issue.

Found by @hhovsepy

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -1463,7 +1463,15 @@ class BootstrapSwitch(BaseInput):
 
     @property
     def selected(self):
-        return self.browser.is_selected(self)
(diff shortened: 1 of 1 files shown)


hhovsepy:
@izapolsk thank you for the fix,
@mfalesni looking forward to see it released, we are blocked by this

mfalesni:
@hhovsepy it is :) https://pypi.python.org/pypi/widgetastic.patternfly

PR #15, opened 2017-08-10
Accordion: Fix header click and harden open

mfalesni:
@psav @mshriver this should resolve the issues we are seeing.

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -16,7 +16,7 @@ from widgetastic.widget import BaseInput, ClickableMixin, TextInput, Text, Widge
     do_not_read_this_widget
 from widgetastic.xpath import quote
 
-from wait_for import wait_for, wait_for_decorator
+from wait_for import wait_for, wait_for_decorator, TimedOutError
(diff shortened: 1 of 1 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -669,17 +670,30 @@ def is_opened(self):
...
+                wait_for(lambda: self.is_opened, delay=0.1, num_sec=3)
+            except TimedOutError:
+                self.logger.warning('Could not open the accordion, trying clicking again')
+                # Workaround stupid pages, perhaps we put a try mechanism in here
+                if self.is_closed:
+                    self.click()
+                    try:
+                        wait_for(lambda: self.is_opened, delay=0.1, num_sec=3)
+                    except TimedOutError:
+                        self.logger.error('Could not open the accordion')
```
mshriver: Consider using logger.exception here so that we get the TimedOutError in the log, and this is the last time its caught.

mfalesni: I did this specifically to not have that exception printed out, because we know why that happens.

PR #18, opened 2017-08-17
Enable partial matching for BootstrapSelect

mfalesni:


lkhomenk:
@mfalesni  same is required for WT.Dropdown

src/widgetastic_patternfly/__init__.py
```diff
@@ -732,6 +732,16 @@ class BootstrapSelect(Widget, ClickableMixin):
     LOCATOR_START = './/div[contains(@class, "bootstrap-select")]'
     ROOT = ParametrizedLocator('{@locator}')
     BY_VISIBLE_TEXT = './div/ul/li/a[./span[contains(@class, "text") and normalize-space(.)={}]]'
+    BY_PARTIAL_VISIBLE_TEXT = (
+        './div/ul/li/a[./span[contains(@class, "text") and contains(normalize-space(.), {})]]')
+
+    class partial(object):  # noqa
```
mshriver: Consider moving this outside of BootstrapSelect class definition so that we can use it for Dropdown and other widgets where we eventually want partial matching.

mfalesni: Good idea.

mfalesni: I created https://github.com/RedHatQE/widgetastic.core/pull/60 , please review that one and then I will use that class for matching.

PR #19, opened 2017-08-23
[RFR]SSUIDropdown added

sshveta:
Most of the code is same as DropDown but the locators are different.


mfalesni:
This is a specific version that is used in ManageIQ that will be merged inside ManageIQ integration tests so closing it here. Thanks for contribution.

PR #23, opened 2017-09-18
Fixed BootstrapTreeview for CFME 5.9

quarckster:
Upstream has changed BootstrapTreeview. Now it has `miq-tree-view` in the root. Tested on CFME 5.7, 5.8 and upstream.

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -644,6 +644,7 @@ class Accordion(View, ClickableMixin):
         './/div[contains(@class, "panel-group")]/div[contains(@class, "panel") and '
         './div/h4/a[normalize-space(.)={@accordion_name|quote}]]')
     TREE_LOCATOR = '|'.join([
+        './/miq-tree-view',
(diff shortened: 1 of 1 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -843,9 +844,12 @@ class BootstrapTreeview(Widget):
     You don't have to specify the ``tree_id`` if the hosting object implements ``tree_id``.
 
     Args:
-        tree_id: Id of the tree, the closest div to the root ``ul`` element.
+        tree_id: Id of the tree, the closest div or ``miq-tree-view`` to the root ``ul`` element.
     """
-    ROOT = ParametrizedLocator('#{@tree_id}')
+    ROOT = ParametrizedLocator('|'.join([
+        './/miq-tree-view[@name={@tree_id}]/div',
```
mshriver: I believe this `tree_id` needs to be quoted here.

Failing for me in testing on miq build from master.20170918, and searching in the xpath this ROOT lookup fails without the value being quoted.

PR #24, opened 2017-09-19
Adding a new Button as required in CFME timeprofile page.

anurag03:
This Save button is required for checking the disabled option of button on "Add page of timeprofile" in CFME.

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -122,7 +122,7 @@ class Button(Widget, ClickableMixin):
 
     @property
     def disabled(self):
-        return self.browser.get_attribute('disabled', self) == 'disabled'
+        return 'disabled' in self.browser.classes(self)
 
     def __repr__(self):
         return '{}{}'.format(type(self).__name__, call_sig(self.args, self.kwargs))
@@ -130,21 +130,6 @@ class Button(Widget, ClickableMixin):
     @property
     def title(self):
         return self.browser.get_attribute('title', self)
-
-
-class SaveButton(Button):
-    """
-        This widget is use for the Save/Add button in CFME
-
-        .. code-block:: python
-            SaveButton('Text of the Button')
-
-        Returns:
-             True if Button is disabled else False
-    """
-    @property
-    def disabled(self):
-        return 'disabled' in self.browser.classes(self)
 
 
 class ViewChangeButton(Widget, ClickableMixin):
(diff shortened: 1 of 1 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -132,6 +132,21 @@ def title(self):
         return self.browser.get_attribute('title', self)
 
 
+class SaveButton(Button):
```
quarckster: I wouldn't create a separate class for that. Just enough to change `disabled` property in the parent class:
```python
@property
def disabled(self):
    return self.browser.get_attribute('disabled', self) == 'disabled' or 'disabled' in self.browser.classes(self)
```

anurag03: I made a new class because I think some of the button will work correctly with disabled property of Button and some will use SaveButton's property, like I am using it in one of my test case.

mshriver: I'd agree with @quarckster here, we don't need a separate class for the disabled property, we'd be much better served by an improved disabled property in the existing button class.

anurag03: Okay @quarckster @mshriver will implement the suggested changes. I didn't implement it on the first place because I thought that there will some buttons which have disabled property working properly.

src/widgetastic_patternfly/__init__.py
```diff
@@ -122,7 +122,7 @@ def active(self):
 
     @property
     def disabled(self):
-        return self.browser.get_attribute('disabled', self) == 'disabled'
+        return 'disabled' in self.browser.classes(self)
```
mfalesni: Shouldn't this check both of these things then? Why are you removing the old conditions? I believe it should be `or`ed

anurag03: okay will make the changes accordingly.

anurag03: This was the reason I tried to write a new button, totally forgot that I can use 'or'

src/widgetastic_patternfly/__init__.py
```diff
@@ -122,7 +122,8 @@ def active(self):
 
     @property
     def disabled(self):
-        return self.browser.get_attribute('disabled', self) == 'disabled'
+        return 'disabled' in self.browser.classes(self) or (self.browser.get_attribute
+                                                            ('disabled', self) == 'disabled')
```
mfalesni: the indentation here iis ugly, please, end the line with or and put the second condition on its line as whole :)

anurag03: Done @mfalesni

PR #25, opened 2017-10-16
Enable tab ignoring

mfalesni:


src/widgetastic_patternfly/__init__.py
```diff
@@ -593,23 +593,40 @@ class Tab(View, ClickableMixin):
...
+            return True
+
     def is_active(self):
         return 'active' in self.browser.classes(self)
 
     def is_disabled(self):
         return 'disabled' in self.browser.classes(self)
 
     def select(self):
+        if self.IGNORE_WHEN_NOT_PRESENT:
```
RonnyPfannschmidt: at first glance this gives an feeling of allowing code to easily look working

for better understanding this should come with a clear note and a example at least

mfalesni: No, it will not. This is to ease the upgrade if old version did not have tabs and new has tabs or vice versa. Remember that by default, widgetastic checks for visibility, so the things "contained" in a tab will yell anyway if something is wrong :)

jkrocil: What's the difference between `present` and `is_displayed`?

mfalesni: Right. My brain processes after weekend. I will remove present :)

mfalesni:
Subclassing can solve this very simply.

PR #41, opened 2018-01-12
[RFR]removed dot from PARTIAL match locator

sshveta:
New 5.9 build is not selecting dropdown values from partial_match(vlan) .
@mfalesni : please review.
It works with 5.8 build too - I have tested for my test .


izapolsk:
duplicate to #40

PR #43, opened 2018-01-23
Update CheckableBootstrapTreeview.fill

### LINKED ISSUE #42: CheckableBootstrapTreeview.fill() needs to uncheck
https://github.com/RedHatQE/widgetastic.patternfly/blob/master/src/widgetastic_patternfly/__init__.py#L1380

This fill method will only check nodes, and won't uncheck those that are filled.

I would like to enhance the fill by taking a more complex data structure with tuples of path and `check` boolean to check/uncheck within a fill context.

The current implementation is forcing callers to use check_node/uncheck_node in patterns that would be better served within a `fill()`

mshriver:
Support check and uncheck in the fill
class scoped namedtuple for CheckNode and UncheckNode action

Tested against ManageIQ/Integration_tests uses of the CheckableBootstrapTreeview fill.

Unit tests are being authored in a separate PR since there were no existing tests.

FIXES #42

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -1346,6 +1346,9 @@ class CheckableBootstrapTreeview(BootstrapTreeview):
     IS_CHECKABLE = './span[contains(@class, "check-icon")]'
     IS_CHECKED = './span[contains(@class, "check-icon") and contains(@class, "fa-check-square-o")]'
 
+    CheckNode = namedtuple('CheckNode', ['path'])
+    UncheckNode = namedtuple('UncheckNode', ['path'])
+
     def is_checkable(self, item):
         return bool(self.browser.elements(self.IS_CHECKABLE, parent=item))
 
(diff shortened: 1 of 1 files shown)


testing/testing_page.html
```diff
@@ -63,7 +65,86 @@ <h1>Widgetastic About Modal</h1>
       </div>
     </div>
   </div>
+
+  <div>
```
mshriver: Wasn't able to get the checkable part of this tree working correctly.

I've tested in the CFME application's implementation of the checkable bootstrap tree, but haven't unit tested via this testing_page.

mfalesni: Is this correct? I would assume you give the JS an empty div and a dictionary of data and it will populate the DOM for you.

mshriver: No, but you are. I pulled the reference from Patternfly's site instead of trying to pull from MIQ.  Working better now.

mshriver:
The unit tests using the updates to testing_page are not identifying the paths/nodes correctly.

In use against views from ManageIQ/integration_tests, this fill works correctly. I believe the CheckableBootstrapTreeview implementation is reviewable, but the testing_page and unit tests need work.

mshriver:
Travis fails on the about_modal test, unrelated.

PR #50, opened 2018-03-15
Remove ROOT from the Tab class

mfalesni:


src/widgetastic_patternfly/__init__.py
```diff
@@ -616,26 +616,33 @@ def __repr__(self):
...
 
 
-class Tab(View, ClickableMixin):
+class Tab(View):
     """Represents the Tab widget.
 
     Selects itself automatically when any child widget gets accessed, ensuring that the widget is
     visible.
+
+    You can specify your own ``ROOT`` attribute on the class.
```
abalakh: I'm not sure about CF, but for sat specifying ROOT will break `TAB_LOCATOR` so tab locator should be specified as well.
Please take a look at your UI, but for us tab link and actual tab content are located in different parts of DOM, so `TAB_LOCATOR` should not rely on ROOT. If that's not the case for CF - then your solution is ok and we'll just provide our specific locators, but if that's the case for CF as well - then the solution is not universal enough and we'd rather think of other options.

mfalesni: I think that when I use parent browser instead of browser, then that should work

src/widgetastic_patternfly/__init__.py
```diff
@@ -660,6 +660,61 @@ def __repr__(self):
...
+            parent=parent)
+
+    def child_widget_accessed(self, widget):
+        """Nothing. Since we don't know which sub_item."""
+
+    def __repr__(self):
+        return '<TabWithDropdown {!r}>'.format(self.tab_name)
+
+
+class TabWithDropdownDefault(TabWithDropdown):
```
abalakh: How about  a bit different naming, like TabWithDropdown/TabWithDropdownDefault -> GenericTabWithDropdown/TabWithDropdown or smth else? It's just sounds very unlikely to me that user will commonly use the class where you have to specify sub_item for each `select()` call, so i'd use "TabWithDropdown" name for the one which will actually be used - it seems kinda weird to use suffix 'Default' every time.

mfalesni: Sure, I can rename them.

PR #51, opened 2018-03-20
Add text partial match for sidebar navigation

rlbabyuk:
Added text partial match for sidebar navigation, as most of them, have some item count as random

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -426,6 +426,7 @@ class BootstrapNav(Widget):
     ROOT = ParametrizedLocator('{@locator}')
     CURRENTLY_SELECTED = './/li[contains(@class, "active")]/a'
     TEXT_MATCHING = './/li/a[text()={txt}]'
+    PARTIAL_TEXT = './/li/a[contains(normalize-space(.), {txt})]'
     ATTR_MATCHING = './/li/a[@{attr}={txt}]'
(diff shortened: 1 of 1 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -454,16 +455,19 @@ def select(self, text=None, **kwargs):
         """Select/click an item from the menu"""
         if text:
             # Select an item based on the text of that item
-            xpath = self.TEXT_MATCHING.format(txt=quote(text))
+            try:
+                link = self.browser.element(self.TEXT_MATCHING.format(txt=quote(text)))
+            except NoSuchElementException:
+                link = self.browser.element(self.PARTIAL_TEXT.format(txt=quote(text)))
```
mfalesni: I believe we should make the choice explicit.

rlbabyuk: done

src/widgetastic_patternfly/__init__.py
```diff
@@ -451,19 +452,30 @@ def read(self):
...
+
+            Args:
+                text: text of the link to be selected, If you want to partial text match,
+                 use the :py:class:`BootstrapNav.partial` to wrap the value.
+        """
         if text:
             # Select an item based on the text of that item
-            xpath = self.TEXT_MATCHING.format(txt=quote(text))
+            if isinstance(text, partial_match):
+                link = self.browser.element(self.PARTIAL_TEXT.format(txt=quote(text)))
```
mfalesni: You should pull the text out of partial match object to be safe.

PR #53, opened 2018-04-13
[RFR]warning should not count as error

digitronik:
In `assert_no_error()` method, flash message type `warning` should not count as error.
for more clarification adding one example[0]

[0] http://pastebin.test.redhat.com/575839

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -221,7 +221,7 @@ class FlashMessages(Widget):
     def assert_no_error(self):
         self.logger.info('asserting there are no error messages')
         for message in self.messages:
-            if message.type not in {'success', 'info'}:
+            if message.type not in {'success', 'info', 'warning'}:
                 self.logger.error('%s: %r', message.type, message.text)
                 raise AssertionError('assert_no_error: {}: {}'.format(message.type, message.text))
             else:


digitronik:
@mfalesni @izapolsk @psav ^^)

src/widgetastic_patternfly/__init__.py
```diff
@@ -221,7 +221,7 @@ def dismiss(self):
     def assert_no_error(self):
         self.logger.info('asserting there are no error messages')
         for message in self.messages:
-            if message.type not in {'success', 'info'}:
+            if message.type not in {'success', 'info', 'warning'}:
```
rlbabyuk: Not sure this is a good idea just to skip warning messages, suggest not to raise Assertion here, but maybe log this message if it's type warning
@psav @mshriver @quarckster what do you think?

quarckster: @digitronik  In the provided example it looks reasonable to exclude warning for here, but I'm not sure about other cases. It might be better to assert a certain flash message:
```python
view.flash.assert_message('Attaching Cloud Volume "cu-24x7" was cancelled by the user')
```

digitronik: @quarckster: you are right, but my point is `assert_no_error()` should check message throws the error or not. Yes, `assert_message()` will work for all type of message as its checking for text. As per my view, I do not agree to treat `warning` message as an error. Method name (`assert_no_error`) itself specifies that it's must have to check only for error. 

`warning` type message is an informative message. We are excluding `info` type message so why we are treating warning message as an error message?

mshriver: I would agree with @digitronik here, that our `assert_no_error` method should only raise exceptions on `error` type messages.

digitronik: @izapolsk whats you opinion on it?

izapolsk: I agree with you and mshriver

PR #55, opened 2018-05-09
DatePicker Widget

digitronik:
Purpose or Intent
=================
- DatePicker widget for selection of date for read-only and read-write date box.
- Supporting `fill` as default method

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -7,6 +7,7 @@ import six
 import time
 from cached_property import cached_property
 from collections import namedtuple
+from datetime import datetime
 
 from widgetastic.exceptions import NoSuchElementException, UnexpectedAlertPresentException, \
     WidgetOperationFailed, StaleElementReferenceException
@@ -1804,3 +1805,118 @@ class AboutModal(Widget):
             # value will include the label from the <strong> block, parse it out
             items.update({key: element_text.replace(key, '', 1).lstrip()})
         return items
+
+
+class DatePicker(Widget):
+    """Represents the Patternfly/Bootstrap DatePicker.
+    It helps to select a date for the read-only box as well as fill date for the read-write box.
+
+      Args:
+        name: name of DatePicker
+        id: id of DatePicker
+
+    .. code-block:: python
+
+       date = DatePicker(name='miq_date_1')
+    """
+    ROOT = ParametrizedLocator(".//*[contains(@class, 'datepicker-dropdown')]")
+    TEXT_BOX = ".//*[@name='{name}' or @id='{id}']"
+    DATE = ".//*[contains(@class, 'datepicker-days')]/table/tbody/tr/*"
+    MONTH = ".//*[contains(@class, 'datepicker-months')]/table/tbody/tr/td/*"
+    YEAR = ".//*[contains(@class, 'datepicker-years')]/table/tbody/tr/td/*"
+    PREV = ".//*[contains(@class, 'prev')]"
+    NEXT = ".//*[contains(@class, 'next')]"
+    DATE_PICKER_SWITH = ".//*[contains(@class, 'datepicker-switch')]"
+
+    def __init__(self, parent, name=None, id=None, logger=None):
+        """Create the widget"""
+        Widget.__init__(self, parent, logger=logger)
+        self.name = name
+        self.id = id
+        self.TEXT_BOX = self.TEXT_BOX.format(name=name, id=id)
+
+    @property
+    def value(self):
+        return self.parent_browser.get_attribute('value', self.TEXT_BOX)
+
+    @property
+    def date_switch_value(self):
+        return self.browser.get_attribute('textContent', self.DATE_PICKER_SWITH)
+
+    def _select_date(self, dd):
+        for el in self.browser.elements(self.DATE):
+            if (not bool({'old', 'new'} & self.browser.classes(el))) and (int(el.text) == int(dd)):
+                el.click()
+                return True
+
+    def _select_month(self, month):
+        for el in self.browser.elements(self.MONTH):
+            if el.text == month:
+                el.click()
+                return True
+
+    def _select_year(self, year):
+        for yr in range(1, 10):
+            start_yr, end_yr = [int(item) for item in self.date_switch_value.split('-')]
+            if start_yr <= int(year) <= end_yr:
+                for el in self.browser.elements(self.YEAR):
+                    if int(el.text) == int(year):
+                        el.click()
+                        return True
+            elif int(year) < start_yr:
+                self.browser.click(self.PREV)
+            elif int(year) > end_yr:
+                self.browser.click(self.NEXT)
+        else:
+            raise ValueError("Year not valid")
+
+    def read(self):
+        return self.value
+
+    def select(self, date):
+        """select date
+
+        Args:
+            date: date in `mm/dd/yyyy` format
+        Returns:
+            :py:class:`bool`
+        """
+        if date == self.value:
+            return False
+
+        self.parent_browser.click(self.TEXT_BOX)
+        date = datetime.strptime(date, "%m/%d/%Y")
+        month = date.strftime("%b")
+        day = date.strftime("%d")
+        year = date.strftime("%Y")
+        sw_value = self.date_switch_value.split(' ')
+
+        if (date.strftime("%B") not in sw_value) or (year not in sw_value):
+            self.browser.click(self.DATE_PICKER_SWITH)
+            if year not in self.date_switch_value:
+                self.browser.click(self.DATE_PICKER_SWITH)
+                self._select_year(year)
+            self._select_month(month)
+        self._select_date(day)
+        return True
+
+    def fill(self, date):
+        """Fill date to date box
+        This method usefull for read-write date box
+
+        Args:
+            date: date in respective format
+        Returns:
+            :py:class:`bool`
+        """
+        if date == self.value:
+            return False
+
+        self.parent_browser.click(self.TEXT_BOX)
+        self.parent_browser.clear(self.TEXT_BOX)
+        self.parent_browser.send_keys(date, self.TEXT_BOX)
+        return True
+
+    @property
+    def is_displayed(self):
+        return self.parent_browser.is_displayed(self.TEXT_BOX)
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -1858,26 +1858,33 @@ class DatePicker(View):
       Args:
         name: Name of DatePicker
         id: Id of DatePicker
-        locator: If none of above apply, you can also supply a full locator
-        strptime_format: `datetime` module `strptime` format. The default is for `mm/dd/yyyy` but 
-        the user can overwrite as per widget requirement, which should comparable with datetime.
+        locator: If none of the above applies, you can also supply a full locator
+        strptime_format: `datetime` module `strptime` format. The default is for `mm/dd/yyyy` but
+        the user can overwrite as per widget requirement which should comparable with datetime.
 
     .. code-block:: python
+        date = DatePicker(name='miq_date_1')
 
-       date = DatePicker(name='miq_date_1')
+        # check readonly or editable
+        date.is_readonly
+        # fill current date
+        date.fill(datetime.now())
+        # read selected date for DatePicker
+        date.read()
     """
     textbox = TextInput(locator=Parameter('@locator'))
 
-    def __init__(self, parent, id=None, name=None, strptime_format='%m/%d/%Y',
-                locator=None, logger=None):
+    def __init__(self, parent, id=None, name=None, locator=None,
+                strptime_format='%m/%d/%Y', logger=None):
         View.__init__(self, parent=parent, logger=logger)
 
         self.strptime_format = strptime_format
+        base_locator = './/*[(self::input or self::textarea) and @{}={}]'
 
         if id:
-            self.locator = './/*[(self::input or self::textarea) and @id={}]'.format(quote(id))
+            self.locator = base_locator.format('id', quote(id))
         elif name:
-            self.locator = './/*[(self::input or self::textarea) and @name={}]'.format(quote(name))
+            self.locator = base_locator.format('name', quote(name))
         elif locator:
             self.locator = locator
         else:
@@ -1904,41 +1911,41 @@ class DatePicker(View):
     @View.nested
     class date_pick(HeaderView):    # noqa
         DATES = ".//*[contains(@class, 'datepicker-days')]/table/tbody/tr/td"
-        _dates = {}
 
         @property
         def _elements(self):
+            dates = {}
             for el in self.browser.elements(self.DATES):
                 if not bool({'old', 'new', 'disabled'} & self.browser.classes(el)):
-                    self._dates.update({int(el.text): el})
-            return self._dates
+                    dates.update({int(el.text): el})
+            return dates
 
     @View.nested
     class month_pick(HeaderView):   # noqa
         MONTHS = ".//*[contains(@class, 'datepicker-months')]/table/tbody/tr/td/*"
-        _months = {}
 
         @property
         def _elements(self):
+            months = {}
             for el in self.browser.elements(self.MONTHS):
                 if not bool({'disabled'} & self.browser.classes(el)):
-                    self._months.update({el.text: el})
-            return self._months
+                    months.update({el.text: el})
+            return months
 
     @View.nested
     class year_pick(HeaderView):    # noqa
         YEARS = ".//*[contains(@class, 'datepicker-years')]/table/tbody/tr/td/*"
-        _years = {}
 
         @property
         def _elements(self):
+            years = {}
             for el in self.browser.elements(self.YEARS):
                 if not bool({'old', 'new', 'disabled'} & self.browser.classes(el)):
-                    self._years.update({int(el.text): el})
-            return self._years
+                    years.update({int(el.text): el})
+            return years
 
         def select(self, value):
-            for yr in range(0, 10):
+            for _yr in range(10):
                 start_yr, end_yr = [int(item) for item in self.datepicker_switch.read().split('-')]
                 if start_yr <= value <= end_yr:
                     for el, web_el in self._elements.items():
@@ -1972,8 +1979,9 @@ class DatePicker(View):
         Returns:
             :py:class:`bool`
         """
-        if self.read():
-            if value.date() == self.read().date(): return False
+        current_date = self.read()
+        if current_date:
+            if value.date() == current_date.date(): return False
 
         if not self.is_readonly:
             date = datetime.strftime(value, self.strptime_format)
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -1952,10 +1952,10 @@ class DatePicker(View):
         def select(self, value):
             start_yr, end_yr = [int(item) for item in self.datepicker_switch.read().split('-')]
             if value > end_yr:
-                for _yr in range(end_yr, value, 10):
+                for _ in range(end_yr, value, 10):
                     self.next_button.click()
             elif value < start_yr:
-                for _yr in range(start_yr, value, -10):
+                for _ in range(start_yr, value, -10):
                     self.prev_button.click()
             self._pick(value)
 
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -12,7 +12,7 @@ from datetime import datetime
 from widgetastic.exceptions import NoSuchElementException, UnexpectedAlertPresentException, \
     WidgetOperationFailed, StaleElementReferenceException
 from widgetastic.log import call_sig
-from widgetastic.utils import ParametrizedLocator, VersionPick, partial_match
+from widgetastic.utils import ParametrizedLocator, Parameter, VersionPick, partial_match
 from widgetastic.widget import BaseInput, ClickableMixin, TextInput, Text, Widget, View, \
     do_not_read_this_widget
 from widgetastic.xpath import quote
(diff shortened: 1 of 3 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -1804,3 +1805,118 @@ def items(self):
...
+        if (date.strftime("%B") not in sw_value) or (year not in sw_value):
+            self.browser.click(self.DATE_PICKER_SWITH)
+            if year not in self.date_switch_value:
+                self.browser.click(self.DATE_PICKER_SWITH)
+                self._select_year(year)
+            self._select_month(month)
+        self._select_date(day)
+        return True
+
+    def fill(self, date):
```
lkhomenk: REQUIRED: please recheck this method. it won't fill-in any values. 
```
<input type="text" name="miq_date_1" id="miq_date_1" value="" readonly="readonly" data-date-orientation="bottom" data-miq_observe_date="{&quot;url&quot;:&quot;/ops/cu_repair_field_changed&quot;}" data-provide="datepicker" data-date-autoclose="true" data-date-format="mm/dd/yyyy" data-date-language="en" data-date-week-start="0">
```
it's read-only

lkhomenk:
if some Datapicker widgets are editable - this should be covered by widget - check/verify if it's editable

digitronik:
@lkhomenk added new methods:
1. `is_readonly` which will return bool 
2. `date_format` which will return date format

src/widgetastic_patternfly/__init__.py
```diff
@@ -1804,3 +1805,130 @@ def items(self):
...
+            return False
+
+        self.parent_browser.click(self.TEXT_BOX)
+
+        if not self.is_readonly:
+            self.parent_browser.clear(self.TEXT_BOX)
+            self.parent_browser.send_keys(date, self.TEXT_BOX)
+            return True
+        else:
+            raise ValueError("readonly datepicker not fillable")
```
lkhomenk: REQUIRED: else self.select should be called

src/widgetastic_patternfly/__init__.py
```diff
@@ -1804,3 +1805,128 @@ def items(self):
...
+            return False
+
+        self.parent_browser.click(self.TEXT_BOX)
+
+        if not self.is_readonly:
+            self.parent_browser.clear(self.TEXT_BOX)
+            self.parent_browser.send_keys(date, self.TEXT_BOX)
+            return True
+        else:
+            self.select(date)
```
lkhomenk: `return` this please

lkhomenk:
locally checked - works as expected.

src/widgetastic_patternfly/__init__.py
digitronik: @lkhomenk done

src/widgetastic_patternfly/__init__.py
digitronik: @mshriver: I added support for `dict` input with keys `day`, `month` and `year` but I  like to take `str` input as well. So now both will accept. 

while writing test, Most of the time user not going to hardcode any date. In python, he will take help of `datetime` module for date manipulation like In my case, I will try to collect 2-3 days back data so I have to create  `datetime.datetime.now()` object and minus how much days I wish to go back. 
```
from datetime import datetime, timedelta
(datetime.today() - timedelta(days=2)).strftime("%m/%d/%y")
Out[44]: '05/16/18'
```
The point is that if I have `datetime` object no need to send `date`, `month` and `year` in `dirt`.  I can convert into `str` as per `format used by the developer` by checking `widget.date_format` property. So for convenience, I added support for both type input.

izapolsk:
@digitronik, travis failed please check what wrong. 
In addition, since this module is shared with other teams, we need to cover our widgets by tests as it is done in widgetastic.core .  Please add appropriate tests.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1804,3 +1805,140 @@ def items(self):
...
+    YEAR = ".//*[contains(@class, 'datepicker-years')]/table/tbody/tr/td/*"
+    PREV = ".//*[contains(@class, 'prev')]"
+    NEXT = ".//*[contains(@class, 'next')]"
+    DATE_PICKER_SWITH = ".//*[contains(@class, 'datepicker-switch')]"
+    DATE_MAPPING = {'dd': '%d', 'mm': '%m', 'MM': '%B', 'yy': '%y', 'yyyy': '%Y'}
+
+    def __init__(self, parent, name=None, id=None, logger=None):
+        """Create the widget"""
+        Widget.__init__(self, parent, logger=logger)
+        self.name = name
```
izapolsk: REQUIRED: this "or" logic should be inside \_\_init\_\_ not TEXT_BOX.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1804,3 +1805,140 @@ def items(self):
...
+      Args:
+        name: name of DatePicker
+        id: id of DatePicker
+
+    .. code-block:: python
+
+       date = DatePicker(name='miq_date_1')
+    """
+    ROOT = ParametrizedLocator(".//*[contains(@class, 'datepicker-dropdown')]")
+    TEXT_BOX = ".//*[@name='{name}' or @id='{id}']"
```
izapolsk: OPTIONAL: I'd be better to call it TEXTBOX then.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1804,3 +1805,140 @@ def items(self):
...
+    DATE = ".//*[contains(@class, 'datepicker-days')]/table/tbody/tr/*"
+    MONTH = ".//*[contains(@class, 'datepicker-months')]/table/tbody/tr/td/*"
+    YEAR = ".//*[contains(@class, 'datepicker-years')]/table/tbody/tr/td/*"
+    PREV = ".//*[contains(@class, 'prev')]"
+    NEXT = ".//*[contains(@class, 'next')]"
+    DATE_PICKER_SWITH = ".//*[contains(@class, 'datepicker-switch')]"
```
izapolsk: REQUIRED: typo

src/widgetastic_patternfly/__init__.py
```diff
@@ -1804,3 +1805,140 @@ def items(self):
...
+    DATE_MAPPING = {'dd': '%d', 'mm': '%m', 'MM': '%B', 'yy': '%y', 'yyyy': '%Y'}
+
+    def __init__(self, parent, name=None, id=None, logger=None):
+        """Create the widget"""
+        Widget.__init__(self, parent, logger=logger)
+        self.name = name
+        self.id = id
+        self.TEXT_BOX = self.TEXT_BOX.format(name=name, id=id)
```
izapolsk: REQUIRED: please use ParametrizedLocator or ParametrizedString instead.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1804,3 +1805,140 @@ def items(self):
...
+
+    @property
+    def value(self):
+        return self.parent_browser.get_attribute('value', self.TEXT_BOX)
```
izapolsk: REQUIRED: good job ! however, I think you have to turn your xpaths above into separate widgets in this widget/view.
Your widget is too dynamic.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1804,3 +1805,140 @@ def items(self):
...
+        else:
+            return self.select(
+                {'day': date_obj.day, 'month': date_obj.month, 'year': date_obj.year})
+
+    @property
+    def is_displayed(self):
+        return self.parent_browser.is_displayed(self.TEXT_BOX)
+
+    @property
+    def is_readonly(self):
```
izapolsk: REQUIRED: all public methods and properties should have docstrings with description

src/widgetastic_patternfly/__init__.py
```diff
@@ -1804,3 +1805,140 @@ def items(self):
...
+        if (date.strftime("%B") not in sw_value) or (year not in sw_value):
+            self.browser.click(self.DATE_PICKER_SWITH)
+            if year not in self.date_switch_value:
+                self.browser.click(self.DATE_PICKER_SWITH)
+                self._select_year(year)
+            self._select_month(month)
+        self._select_date(day)
+        return True
+
+    def fill(self, value):
```
izapolsk: REQUIRED: first of all I believe it should accept datetime instead.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1804,3 +1805,140 @@ def items(self):
...
+                self.browser.click(self.PREV)
+            elif int(year) > end_yr:
+                self.browser.click(self.NEXT)
+        else:
+            raise ValueError("Year not valid")
+
+    def read(self):
+        return self.value
+
+    def select(self, date):
```
izapolsk: QUESTION: what is it for ? there is fill method.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1804,3 +1805,140 @@ def items(self):
...
+                        el.click()
+                        return True
+            elif int(year) < start_yr:
+                self.browser.click(self.PREV)
+            elif int(year) > end_yr:
+                self.browser.click(self.NEXT)
+        else:
+            raise ValueError("Year not valid")
+
+    def read(self):
```
izapolsk: REQUIRED: I believe it should return datetime.

src/widgetastic_patternfly/__init__.py
digitronik: now it's internal method :)

digitronik:
@izapolsk I added suggested changes. Please review it one time and let me know if changes required.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1804,3 +1805,150 @@ def items(self):
...
+        name: name of DatePicker
+        id: id of DatePicker
+
+    .. code-block:: python
+
+       date = DatePicker(name='miq_date_1')
+    """
+    ROOT = ParametrizedLocator(".//*[contains(@class, 'datepicker-dropdown')]")
+    TEXTBOX = ParametrizedLocator(".//*[@name='{@name}' or @id='{@id}']")
+    DATE = ".//*[contains(@class, 'datepicker-days')]/table/tbody/tr/*"
```
izapolsk: REQUIRED: sorry, I probably explained not well enough what I want. All this stuff should be turned into widgets inside main widget. 
This is not the way we develop our widgets. Please look at examples.

src/widgetastic_patternfly/__init__.py
izapolsk: I still see logic inside that TEXTBOX xpath instead of here.

src/widgetastic_patternfly/__init__.py
digitronik: @izapolsk: I added small changes but I am not able to convert all into widgets. I converted some; which were possible. 

`DATE`, `MONTH`, `YEAR` are `table` but it's difficult and hectic task to operate on the element as per `DatePicker` requirement. Here, we required all `cell` elements of the table but  it have special header elements (`<<`, `>>`, and `Switch`); As per `Table` widget `rows` are mapped with `headers` result, getting extra rows with special characters like `<<` header. It's makes sorting difficult and Implementation will increase the line of code. As, we have to some operation on `classes` like checking `active`, `old` etc status. So, I found best to go with all elements.

izapolsk: @digitronik, I'll look over your changes and let's discuss this today.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1849,3 +1850,172 @@ def click_location(self, name, handle_alert=True):
...
+    .. code-block:: python
+
+       date = DatePicker(name='miq_date_1')
+    """
+    textbox = TextInput(locator=Parameter('@locator'))
+    DATE_MAPPING = {'dd': '%d', 'mm': '%m', 'MM': '%B', 'yy': '%y', 'yyyy': '%Y'}
+
+    def __init__(self, parent, id=None, name=None, locator=None, logger=None):
+        View.__init__(self, parent=parent, logger=logger)
+        if id is not None:
```
izapolsk: it will be enough just
`if id: `

digitronik: @izapolsk: I changed it.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1849,3 +1850,172 @@ def click_location(self, name, handle_alert=True):
...
+
+    def fill(self, value):
+        """Fill date to date box
+
+        Args:
+           value (datetime): datetime object.
+        Returns:
+            :py:class:`bool`
+        """
+        if isinstance(self.read(), datetime):
```
izapolsk: I think you can replace all this statement with something like 
`if self.read() == value:
      return False
`

src/widgetastic_patternfly/__init__.py
```diff
@@ -1849,3 +1850,172 @@ def click_location(self, name, handle_alert=True):
...
+            if value.date() == self.read().date():
+                return False
+
+        if not self.is_readonly:
+            date = datetime.strftime(value, self.date_mapp)
+            self.textbox.fill(date)
+            self.date_pick._elements[self.date_pick.active].click()
```
izapolsk: what is it for ?

src/widgetastic_patternfly/__init__.py
```diff
@@ -1849,3 +1850,172 @@ def click_location(self, name, handle_alert=True):
...
+            return True
+        else:
+            self.browser.click(self.textbox)
```
izapolsk: textbox should support click. Why do you click it thru browser ?

src/widgetastic_patternfly/__init__.py
```diff
@@ -1849,3 +1850,172 @@ def click_location(self, name, handle_alert=True):
...
+            self.locator = './/*[normalize-space(@name)={}]'.format(quote(name))
+        elif locator:
+            self.locator = locator
+        else:
+            raise TypeError('You need to specify either, id, name or locator for DatePicker')
+
+    class HeaderView(View):
+        prev_button = Text(".//*[contains(@class, 'prev')]")
+        next_button = Text(".//*[contains(@class, 'next')]")
+        dp_switch = Text(".//*[contains(@class, 'datepicker-switch')]")
```
izapolsk: please give it more clear name

src/widgetastic_patternfly/__init__.py
```diff
@@ -1849,3 +1850,172 @@ def click_location(self, name, handle_alert=True):
...
+                return False
+
+        if not self.is_readonly:
+            date = datetime.strftime(value, self.date_mapp)
+            self.textbox.fill(date)
+            self.date_pick._elements[self.date_pick.active].click()
+            return True
+        else:
+            self.browser.click(self.textbox)
+            sw_value = self.date_pick.dp_switch.read().split(' ')
```
izapolsk: I think you need to just set passed values w/o any parsing. Is there a reason to read those values and check them ?

src/widgetastic_patternfly/__init__.py
```diff
@@ -1849,3 +1850,172 @@ def click_location(self, name, handle_alert=True):
...
+        def select(self, value):
+            for el, web_el in self._elements.items():
+                if el == value:
+                    web_el.click()
+                    return True
+
+        @property
+        def active(self):
+            for el, web_el in self._elements.items():
+                if bool(self.browser.classes(web_el) & {'active', 'focused'}):
```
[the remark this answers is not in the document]
digitronik: Only one of them

src/widgetastic_patternfly/__init__.py
digitronik: texbox is `TextInput` widget. it does not support `click` method directly. So I took help of `browser`

src/widgetastic_patternfly/__init__.py
digitronik: Its bootstrap this which, I found. when we fill textbox, calendar window will open. To close that window better to check for `active` means selected `dd` and click on it.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1849,3 +1850,172 @@ def click_location(self, name, handle_alert=True):
...
+    @View.nested
+    class date_pick(HeaderView):    # noqa
+        DATES = ".//*[contains(@class, 'datepicker-days')]/table/tbody/tr/td"
+        _dates = {}
+
+        @property
+        def _elements(self):
+            for el in self.browser.elements(self.DATES):
+                if not bool({'old', 'new', 'disabled'} & self.browser.classes(el)):
+                    self._dates.update({int(el.text): el})
```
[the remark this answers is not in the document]
digitronik: for `date_pick`, `month_pick` and `year_pick`, `select` and `active` methods are common. date and year values are `int` and others `str`. Also, I required elements for a `click` operation. I created `dict` at `_element` level only to avoid repetitive methods also to avoid repetitive iteration over elements.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1849,3 +1850,172 @@ def click_location(self, name, handle_alert=True):
...
+                            return True
+                elif value < start_yr:
+                    self.prev_button.click()
+                elif value > end_yr:
+                    self.next_button.click()
+            else:
+                raise ValueError("Not valid year or with more than century difference")
+
+    @property
+    def date_mapp(self):
```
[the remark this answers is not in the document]
digitronik: I completely removed this and take directly `strptime_format` form user itself.

src/widgetastic_patternfly/__init__.py
digitronik: Yes, 
First switch value will be a combination of month and Year like `June 2018` so I am splitting it and check for `month` and `year` match or not. if not match it means I have to go for `year` or `month` changes. When we click first time to `switch` its come to month selection view. and its value is a year like `2018`.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1849,3 +1850,172 @@ def click_location(self, name, handle_alert=True):
...
+
+       date = DatePicker(name='miq_date_1')
+    """
+    textbox = TextInput(locator=Parameter('@locator'))
+    DATE_MAPPING = {'dd': '%d', 'mm': '%m', 'MM': '%B', 'yy': '%y', 'yyyy': '%Y'}
+
+    def __init__(self, parent, id=None, name=None, locator=None, logger=None):
+        View.__init__(self, parent=parent, logger=logger)
+        if id is not None:
+            self.locator = './/*[normalize-space(@id)={}]'.format(quote(id))
```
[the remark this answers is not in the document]
digitronik: ok, 
`input` as child so  adding possibility of `(self::input or self::textarea)`

digitronik:
@izapolsk: I answered ques and try to implement suggestions. Please have a look one time.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1849,3 +1850,171 @@ def click_location(self, name, handle_alert=True):
...
+        """Fill date to DatePicker
+
+        Args:
+           value: datetime object.
+
+        Returns:
+            :py:class:`bool`
+        """
+        if self.read():
+            if value.date() == self.read().date(): return False
```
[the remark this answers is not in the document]
digitronik: @izapolsk, I accepted but as `read` return `None` for ` `. Single line check will give `AttributeError` due to `None.date()`. So, here I choose to read in one variable.

I added other changes, Please review it.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1849,3 +1850,179 @@ def click_location(self, name, handle_alert=True):
...
+        @property
+        def _elements(self):
+            years = {}
+            for el in self.browser.elements(self.YEARS):
+                if not bool({'old', 'new', 'disabled'} & self.browser.classes(el)):
+                    years.update({int(el.text): el})
+            return years
+
+        def select(self, value):
+            for _yr in range(10):
```
izapolsk: Why do you add such restriction ? 
As far as I can see widget allows you to choose any value.
Besides you can easily count number of transitions and get to right range. However, you test value after every click. I don't think this is optimal.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1849,3 +1850,179 @@ def click_location(self, name, handle_alert=True):
...
+        """Fill date to DatePicker
+
+        Args:
+           value: datetime object.
+
+        Returns:
+            :py:class:`bool`
+        """
+        current_date = self.read()
+        if current_date:
```
izapolsk: you can combine both if statements

src/widgetastic_patternfly/__init__.py
```diff
@@ -1849,3 +1850,179 @@ def click_location(self, name, handle_alert=True):
...
+            date = datetime.strftime(value, self.strptime_format)
+            self.textbox.fill(date)
+            self.date_pick._elements[self.date_pick.active].click()
+            return True
+        else:
+            self.browser.click(self.textbox)
+            sw_value = self.date_pick.datepicker_switch.read().split(' ')
+            if (value.strftime("%B") not in sw_value) or (str(value.year) not in sw_value):
+                self.date_pick.datepicker_switch.click()
+                if str(value.year) is not self.month_pick.datepicker_switch.read():
```
izapolsk: What do you check here ? "is" is used to check if both vars refer to the same object.

BTW, I doubt that date picker works too slow if you just change value each time.
So, I believe you can remove most these checks above. You need to check whether value and read are different and that's all. If those are different go and pick year, month, date.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1849,3 +1850,179 @@ def click_location(self, name, handle_alert=True):
...
+                    self.month_pick.datepicker_switch.click()
+                    self.year_pick.select(value=value.year)
+                self.month_pick.select(value=value.strftime("%b"))
+            self.date_pick.select(value=value.day)
+            return True
+
+    @property
+    def is_readonly(self):
```
izapolsk: I would just call it readonly. this is property not method.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1849,3 +1850,179 @@ def click_location(self, name, handle_alert=True):
...
+        return self.browser.get_attribute('data-date-format', self.textbox)
+
+    @property
+    def is_displayed(self):
+        """DatePicker displayed or not
+
+        Returns:
+            :py:class:`bool`
+        """
+        return self.parent_browser.is_displayed(self.textbox)
```
izapolsk: why parent_browser ? what isn't ok with self.browser ?

testing/test_date_picker.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def test_bootstrap_date_picker(browser):
+    class TestView(View):
+        dp_readwrite = DatePicker(name="date_readwrite")
+        dp_readonly = DatePicker(id="date_readonly")
+
+    view = TestView(browser)
+
+    # `is_readonly` proprety working or not
```
izapolsk: typo

testing/test_date_picker.py
```diff
@@ -0,0 +1,31 @@
...
+    assert view.dp_readonly.is_readonly
+    assert not view.dp_readwrite.is_readonly
+
+    # `fill` and `read` with current date
+    today_date = datetime.now()
+    view.dp_readonly.fill(today_date)
+    assert today_date.date() == view.dp_readonly.read().date()
```
izapolsk: I'd also test a case when  widget.fill returns False if it is being filled with same date once more

testing/test_date_picker.py
```diff
@@ -0,0 +1,31 @@
...
+
+    view.dp_readwrite.fill(today_date)
+    assert today_date.date() == view.dp_readwrite.read().date()
+
+    # `fill` and `read` with year back date
+    yr_back_date = (today_date - timedelta(days=365))
```
izapolsk: please also test changing month and day

testing/testing_page.html
```diff
@@ -66,6 +66,67 @@ <h1>Widgetastic About Modal</h1>
   </div>
 
 
+<!--------------------------------- Bootstrap-Datepicker----------------------------------------->
+<!-- Special version of Bootstrap that only affects content wrapped in .bootstrap-iso -->
+<link rel="stylesheet" href="https://formden.com/static/cdn/bootstrap-iso12.css" />
```
izapolsk: I doubt all this stuff is neccessary especially in widgitastic.patternfly. Please add two patternfly's datepicker widgets.
Furthermore, all js and css stuff for has already been added to this page above. code examples you can find in manageiq and https://www.patternfly.org/pattern-library/forms-and-controls/datepicker/#code

src/widgetastic_patternfly/__init__.py
digitronik: ok, I like to remove all checks.

testing/test_date_picker.py
digitronik: I like to go with `timedelta` of  `random.randint(100, 5000)`. It will give me an awesome difference.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1849,3 +1849,177 @@ def click_location(self, name, handle_alert=True):
...
+        def _pick(self, value):
+            for el, web_el in self._elements.items():
+                if el == value:
+                    web_el.click()
+                    return True
+
+        def select(self, value):
+            start_yr, end_yr = [int(item) for item in self.datepicker_switch.read().split('-')]
+            if value > end_yr:
+                for _yr in range(end_yr, value, 10):
```
izapolsk: are you sure about amount of transitions ? I haven't checked but it seems it makes one more transition than it should.
as for underscore variables, I meant using just underscore instead of variable name.
you can read more [here](https://hackernoon.com/understanding-the-underscore-of-python-309d1a029edc) about this syntax convention.

digitronik: No transition is fine. I tested it up to a selection of `1900` to `3000`

testing/test_date_picker.py
```diff
@@ -0,0 +1,55 @@
...
+    # check `fill` and `read` with year back date
+    yr_back_date = (today_date - timedelta(days=365))
+    view.dp_readonly.fill(yr_back_date)
+    assert yr_back_date.date() == view.dp_readonly.read().date()
+    view.dp_readwrite.fill(yr_back_date)
+    assert yr_back_date.date() == view.dp_readwrite.read().date()
+    view.dp_basic_patternfly.fill(yr_back_date)
+    assert yr_back_date.date() == view.dp_basic_patternfly.read().date()
+
+    # check if date already selected `fill` should return False
```
mshriver: :+1:

PR #56, opened 2018-05-16
[RFR] Add breadcrumb widget

mshriver:
copypasta from ManageIQ/integration_tests/widgetastic_manageiq

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -1804,3 +1804,48 @@ class AboutModal(Widget):
             # value will include the label from the <strong> block, parse it out
             items.update({key: element_text.replace(key, '', 1).lstrip()})
         return items
+
+
+class BreadCrumb(Widget):
+    """ Patternfly BreadCrumb navigation control
+
+    .. code-block:: python
+
+        breadcrumb = BreadCrumb()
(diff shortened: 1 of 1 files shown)


mshriver:
Travis failure is from about modal in py2.7, not related.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1804,3 +1804,48 @@ def items(self):
...
+
+class BreadCrumb(Widget):
+    """ Patternfly BreadCrumb navigation control
+
+    .. code-block:: python
+
+        breadcrumb = BreadCrumb()
+        breadcrumb.click_location(breadcrumb.locations[0])
+    """
+    ROOT = '//ol[@class="breadcrumb"]'
```
abalakh: May i suggest slightly different locator?
`//ol[contains(@class, "breadcrumb")]`?
Breadcrumb in satellite just have 2 classes, so we'll have to override it (which is still ok for us)

mshriver: Double checking MIQ breadcrumbs to know the scope, but sounds like a great improvement.

mshriver:
I don't have merge rights here, yet, @abalakh, we'll have to get @psav to merge I believe.

PR #69, opened 2018-09-06
Update travis config for autodeploy with mshriver

mshriver:
I need to be added as maintainer on pypi before this will work.


mshriver:
gotta merge it to test it :/

PR #70, opened 2018-09-18
[RFR] Add AggregateStatusCard widget and tests

mshriver:
Adding widgets for Aggregate Status cards, per patternfly reference markup.

Making the locators modular, so that modifying them in application-specific widgets is easier.

PR #71, opened 2018-09-18
Add PFIcon class

mshriver:
Looking for some early feedback about design before adding more icon strings and making changes to other widgets or page modeling/testing.

The intent of this class is to capture the patternfly icon strings into class constants, that we can then use in tests instead of string fragments.

I've included this util in PR #70, and used it in the new widget there, with tests, in order to demonstrate the intended classmethod use.

Also updating travis config to include firefox latest-esr, and updating conftest browser fixture to use headless firefox.


testing/conftest.py
```diff
@@ -22,7 +20,9 @@ def product_version(self):
 
 @pytest.fixture(scope='session')
 def selenium(request):
-    driver = webdriver.PhantomJS()
+    options = webdriver.FirefoxOptions()
+    options.add_argument('-headless')
```
quarckster: **OPTIONAL**
Selenium throws a warning here, this line doesn't:
```python
options.headless = True
```

psav: Yeh I htink that's a nice addition

mshriver: Good call @quarckster looks like its a preferred/deprecated use.

PR #76, opened 2018-12-04
[RFR] fix flash messages in hammer

izapolsk:
current flash widget cannot handle cases when each message is stored in its own flash_text_div. this fix should sort out this issue

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -8,6 +8,7 @@ import time
 from cached_property import cached_property
 from collections import namedtuple
 from datetime import datetime
+from operator import not_
 
 from widgetastic.exceptions import (
     NoSuchElementException,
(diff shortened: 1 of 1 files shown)


mshriver:
Testing in MIQ/integration_tests is looking good.

src/widgetastic_patternfly/__init__.py
```diff
@@ -227,39 +230,58 @@ def dismiss(self):
         for message in self.messages:
             message.dismiss()
 
-    def assert_no_error(self):
-        self.logger.info('asserting there are no error messages')
+    def match_messages(self, text=None, t=None, partial=False, inverse=False):
+        msg_type = t if isinstance(t, (tuple, list, set)) or t is None else (t, )
```
mshriver: You can simplify this check a bit by using `type(None)` so that it reads:

`msg_type = t if isinstance(t, (tuple, list, set, type(None))) else (t, )`

src/widgetastic_patternfly/__init__.py
```diff
@@ -227,39 +230,58 @@ def dismiss(self):
         for message in self.messages:
             message.dismiss()
 
-    def assert_no_error(self):
-        self.logger.info('asserting there are no error messages')
+    def match_messages(self, text=None, t=None, partial=False, inverse=False):
+        msg_type = t if isinstance(t, (tuple, list, set, type(None))) else (t, )
+        op = bool if inverse else not_
+        log_part = 'partial' if partial else ''
```
mshriver: consider using `'exact'` here for `else`

src/widgetastic_patternfly/__init__.py
```diff
@@ -227,39 +230,58 @@ def dismiss(self):
...
-                self.logger.error('%s: %r', message.type, message.text)
-                raise AssertionError('assert_no_error: {}: {}'.format(message.type, message.text))
-            else:
-                self.logger.info('%s: %r', message.type, message.text)
-
-    def assert_message(self, text, t=None, partial=False):
-        log_part = 'partial match of' if partial else ''
+    def match_messages(self, text=None, t=None, partial=False, inverse=False):
+        msg_type = t if isinstance(t, (tuple, list, set, type(None))) else (t, )
+        op = bool if inverse else not_
```
mshriver: Discussed an in-line comment here, because I actually read this as exactly the opposite of what it does, and then was confused by the logic in the blocks below.

Its the right choice of implementation, just needs some comments to clarify.

src/widgetastic_patternfly/__init__.py
```diff
@@ -227,39 +230,58 @@ def dismiss(self):
...
+                           (not partial and text == message.text)):
+                continue
+
+            matched_messages.append(message)
+        return matched_messages
+
+    def assert_no_error(self, wait=0):
+        self.logger.info('asserting there are no error messages')
+        try:
+            msgs = wait_for(self.match_messages, func_kwargs=dict(t={'success', 'info', 'warning'},
```
mshriver: optional:
unpack the tuple on assignment instead of indexing the wait_for return:

`msgs, _ = wait_for(stuff)`

src/widgetastic_patternfly/__init__.py
```diff
@@ -227,39 +230,61 @@ def dismiss(self):
...
-                                 .format(log_part, e_text, [msg.text for msg in all_messages]))
+            raise AssertionError('assert {} match of message: {}. \n Available messages: {}'
+                                 .format(log_part, e_text, [msg.text for msg in self.messages]))
 
-    def assert_success_message(self, text, t=None, partial=False):
-        self.assert_no_error()
-        self.assert_message(text, t='success', partial=partial)
+    def assert_success_message(self, text, t=None, partial=False, wait=0):
+        self.assert_no_error(wait=wait)
+        self.assert_message(text, t=t or 'success', partial=partial, wait=wait)
```
mshriver: :+1: guess we were ignoring `t` before, good change.

src/widgetastic_patternfly/__init__.py
```diff
@@ -227,39 +230,61 @@ def dismiss(self):
...
+            return True
+
+    def assert_message(self, text, t=None, partial=False, wait=0):
+        log_part = 'partial' if partial else 'exact'
+        try:
+            msgs = wait_for(self.match_messages, func_kwargs=dict(text=text, t=t, partial=partial),
+                            timeout=wait)[0]
+            if msgs:
+                return True
+            else:
```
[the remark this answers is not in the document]
mshriver: Ignore me, I wrongly assumed wait_for would not raise TimedOutError on a `0` timeout. Gotta account for that, and raising here just in case is good.

Issue #75, opened 2018-11-28
Kebab Widget from miq


digitronik:
Need to add `kebab` menu widget from `miq` with proper tests. It's patternfly` widget.

https://www.patternfly.org/pattern-library/widgets/#kebabs

digitronik:
done

PR #77, opened 2018-12-07
[RFR] Kebab  Widget

digitronik:
- Moving kebab widget from [miq](https://github.com/ManageIQ/integration_tests/blob/master/cfme/dashboard.py#L23)
- Changes:
    - Added `locator` as on of optional agrument. So if id not match and want to select `kebab` other than first. 
    - As its `dropdown` type  of  widget so methods should  similar to dropdown. renamed `select` method with `item_select`.
    - Added `has_item` method

- Added test

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -2271,8 +2271,10 @@ class Kebab(Widget):
         Widget.__init__(self, parent=parent, logger=logger)
 
         if id:
-            self.locator = './/div[contains(@class, "dropdown-kebab-pf") and ./button[@id={}]]'.format(
-                quote(id)
+            self.locator = (
+                './/div[contains(@class, "dropdown-kebab-pf") and ./button[@id={}]]'.format(
+                    quote(id)
+                )
             )
         elif locator:
             self.locator = locator
@@ -2320,7 +2322,7 @@ class Kebab(Widget):
         Args:
             item: Item to be selected.
             close: Whether to close the kebab after selection. If the item is a link, you may want
-                to set this to ``False``
+            to set this to `False`.
         """
         try:
             el = self.browser.element(self.ITEM.format(quote(item)))
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -2256,12 +2256,11 @@ class Kebab(Widget):
 
     Args:
         id: Id of Kebab button
-        locator: If id not found, you can also supply a full locator
-
-        If nothing specified, first kebab available will be used
+        locator: Kebab button locator
     """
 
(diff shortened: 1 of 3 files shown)


digitronik:
related to https://github.com/RedHatQE/widgetastic.patternfly/issues/75

digitronik:
Indexing is just trial/proposed idea; else I will revert this...

digitronik:
@mshriver: After understanding the concepts of `inheritance` as per your suggestions on RC l.e. In the future if someone like to inherit `Kebab`and try to change locators as per project requirement. 

We are providing the main two options:
1. `id`:
       If a widget has a specific id then we can go with this option. For extension means, if anyone inherited and want to change then a user can just change `BASE_LOCATOR`

2. `locator`:
      `locator` is an important argument as per my observation. Let suppose, we won't have `Id` then we can go with the button locator. I would like every new widget has a locator option.

3. `index`: it's completely `optional`. But, I like to introduce as per current implementation. It not going to effect on main widget usage. It's pointing to `patternfly kebab` `standard implementation`.

trying to explain indexing: `(which is optional)` (With Milan local implementation)
- Suppose; I have a page having only one kebab like `SSUI`:
`x = Kebab(view)`   `# It will give you first kebab on a page. It just like PaginationPlane view`
`Note`: the other two argument options available. the user can use them.

- Suppose; I have three kebabs on the same page: Like `XPath` we can check for `index` while widget initialization. 
`x=Kebab(view, index=3)`

I understand `index` is not the proper way, but we can not say it not a `practical` approach. As `id` or `locator` changes as per development changes. It should be one option. Users can decide with which they like to go.

src/widgetastic_patternfly/__init__.py
```diff
@@ -2249,3 +2249,82 @@ class AggregateStatusMiniCard(AggregateStatusCard):
...
+    """
+
+    ROOT = ParametrizedLocator("{@locator}")
+    BASE_LOCATOR = ".//div[contains(@class, 'dropdown-kebab-pf') and ./button[@id={}]]"
+    UL = './ul[contains(@class, "dropdown-menu")]'
+    BUTTON = "./button"
+    ITEM = "./ul/li/a[normalize-space(.)={}]"
+    ITEMS = "./ul/li/a"
+
+    def __init__(self, parent, id=None, locator=None, index=1, logger=None):
```
mshriver: I don't think this behavior should exist in this widget definition. providing a broad 'locator' parameter covers this use case, but I don't want to encourage that use.

mshriver: The DOM could change, with kebabs removed or inserted, breaking the expectations of the locator defined by index.  There would be no way to tell you're interacting with a different widget than the last one.

digitronik:
@mshriver removed indexing completely... :+1:

Issue #67, opened 2018-08-08
LineChart widget


digitronik:
Need to move `LineChart` widget form `MangeIQ local`[0] to `Patternfly`[1].

I found widget not that much mature, Need to refactor code and fallow `WT.core` guidelines.

[0] https://github.com/ManageIQ/integration_tests/blob/master/widgetastic_manageiq/__init__.py#L4005
[1] https://www.patternfly.org/pattern-library/data-visualization/line-chart/

digitronik:
Please, don't close this issue... `Pie charts` are still in progress.

PR #78, opened 2018-12-15
Charts:  Line, Bar, Spiral widgets

digitronik:
Part of https://github.com/RedHatQE/widgetastic.patternfly/issues/67

Adding fallowing `widgets` with `unit  test`:
```
1. Sparkline
2. Line Chart
    a. single line
    b. line chart
    c. single spline chart
    d. spline chart
3. Bar Chart
   a. Vertical Bar Chart
    a. Grouped Vertical Bar chart
    b. Stacked Vertical Bar Chart
    c. Horizontal Bar Chart
    d. Grouped Horizontal Bar Chart
    e. Stacked Horizontal Bar Chart
```

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -11,16 +11,24 @@ from datetime import datetime
 from operator import not_
 
 from widgetastic.exceptions import (
+    LocatorNotImplemented,
     NoSuchElementException,
+    StaleElementReferenceException,
     UnexpectedAlertPresentException,
     WidgetOperationFailed,
-    StaleElementReferenceException,
-    LocatorNotImplemented,
-    )
+)
 from widgetastic.log import call_sig
-from widgetastic.utils import ParametrizedLocator, Parameter, VersionPick, partial_match
-from widgetastic.widget import BaseInput, ClickableMixin, TextInput, Text, Widget, View, \
-    do_not_read_this_widget
+from widgetastic.utils import Parameter, ParametrizedLocator, VersionPick, partial_match
+from widgetastic.widget import (
+    BaseInput,
+    ClickableMixin,
+    Table,
+    Text,
+    TextInput,
+    View,
+    Widget,
+    do_not_read_this_widget,
+)
 from widgetastic.xpath import quote
 
 from wait_for import wait_for, wait_for_decorator, TimedOutError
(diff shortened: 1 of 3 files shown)


digitronik:
[Notebook](https://drive.google.com/file/d/1qCXX-_xh2S7O1VAucGKkdA8IcJcN-Jgu/view?usp=sharing)  for reviewers... include setup and test.

src/widgetastic_patternfly/__init__.py
```diff
@@ -2274,3 +2282,219 @@ class AggregateStatusMiniCard(AggregateStatusCard):
...
+        """Create the widget"""
+        Widget.__init__(self, parent, logger=logger)
+        if id:
+            self.locator = self.BASE_LOCATOR.format(quote(id))
+        elif locator:
+            self.locator = locator
+        else:
+            raise TypeError("You need to specify either id or locator")
+
+    def read(self, *args, **kwargs):
```
psav: I wonder if in the future we want to move towards using more of a generator approach for reads like this, allowing people to do a performance friendly search operation. This is NOT a requirement for this PR however.

quarckster: @digitronik why do you need these arguments?

src/widgetastic_patternfly/__init__.py
```diff
@@ -2274,3 +2282,219 @@ class AggregateStatusMiniCard(AggregateStatusCard):
...
+
+    Args:
+        id: id of SparkLineChart
+        locator: If id not applies, you can also supply a full locator
+
+    .. code-block:: python
+        spark_line_chart = SparkLineChart(id="sparklineChart")
+    """
+    ROOT = ParametrizedLocator("{@locator}")
+    BASE_LOCATOR = ".//div[@id={}]"
```
mshriver: Please include 'chart-pf-sparkline' class in the default/BASE_LOCATOR to be more explicit here.  The Patternfly spec includes that class. Multiple div's on the page could have the same ID but different classes. (may be a poor page design, but its possible).

src/widgetastic_patternfly/__init__.py
```diff
@@ -2274,3 +2282,219 @@ class AggregateStatusMiniCard(AggregateStatusCard):
...
+        """read all data on chart
+
+        Returns:
+            :py:class:`list` complete data on chart
+        """
+        data = []
+        for el in self.browser.elements(self.RECT):
+            self.browser.move_to_element(el)
+            data.append(self.tooltip.read())
+        return data
```
mshriver: I'm thinking this would be better structured as a dictionary, keyed on the `x` attribute from element matched by `self.RECT`.

src/widgetastic_patternfly/__init__.py
```diff
@@ -2351,3 +2359,219 @@ def item_select(self, item, close=True):
         finally:
             if close:
                 self.close()
+
+
+class SparkLineChart(Widget, ClickableMixin):
+    """Represents the Spark Line Chart from Patternfly (Data Visualization).
+
```
john-dupuy: Providing urls to the Patternfly page for each widget could be helpful.

src/widgetastic_patternfly/__init__.py
```diff
@@ -2351,3 +2359,219 @@ def item_select(self, item, close=True):
...
+        return {self.browser.text(leg): leg for leg in self.browser.elements(self.LEGENDS)}
+
+    @property
+    def legends(self):
+        """ Get all available legends
+
+        Returns:
+            :py:class:`list` all available legends
+        """
+        return list(self._legends.keys())
```
john-dupuy: Why is it necessary to do `list(self._legends.keys())` here? Isn't `self._legends.keys()` already a list?

john-dupuy: Also, why only have a `property` for the text of each legend? What about for the browser elements? e.g.
```
@property 
def legend_text():
    ...
@property
def legend_element():
    ...
```
This may be more readable, but not really necessary.

src/widgetastic_patternfly/__init__.py
```diff
@@ -2351,3 +2359,219 @@ def item_select(self, item, close=True):
...
+
+    def legend_is_displayed(self, leg):
```
john-dupuy: Would it be beneficial to have each legend be it's own class and then have each of these methods be defined within the `Legend` class? I.e. The `legends` property above would return a list of `Legend` objects that each have methods for (`is_displayed`, `hide`, `display`, and `read`). Then these aggregate methods could just act on each `Legend` object.

john-dupuy: This may add too much unnecessary abstraction/code however, so let me know if it doesn't make sense for this particular case.

digitronik: yup; LOC will increase :+1:  which not make any seance. In `WT` we are playing with `web-elements` so better to go with `methods` because legends count depends on `developer` approach. He can add multiple legends each time... we have to need initialize `class` and use methods which will be hectic. I would like to go with `method`.

src/widgetastic_patternfly/__init__.py
digitronik: May be @RonnyPfannschmidt can give me some hack and learning approach towards such handling.

src/widgetastic_patternfly/__init__.py
digitronik: I like this suggestion :100: . Its in my pocket... I want to improve `code-coverage` of `wt.pfy`. I am planning to spend day and push all test for all `widgets`. At same time I will add this for all `widgets`.

src/widgetastic_patternfly/__init__.py
```diff
@@ -2274,3 +2282,219 @@ class AggregateStatusMiniCard(AggregateStatusCard):
...
+    Args:
+        id: id of SparkLineChart
+        locator: If id not applies, you can also supply a full locator
+
+    .. code-block:: python
+        spark_line_chart = SparkLineChart(id="sparklineChart")
+    """
+    ROOT = ParametrizedLocator("{@locator}")
+    BASE_LOCATOR = ".//div[@id={}]"
+    RECT = ".//*[contains(@class, 'c3-event-rects c3-event-rects-single')]//*"
```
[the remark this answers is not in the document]
digitronik: converting to `RECTS`

src/widgetastic_patternfly/__init__.py
RonnyPfannschmidt: on python3 dict keys are indeed a iterables instead of a materialized list
the multi python version support pattern is to indeed wrap keys into a list call

depending on the intended usage-patterns it may beneficial to return a `set(somedict)` instead of a `list(somedict)` - what are those results used for?

digitronik: those are `legends` means `str` elements for graph data pointing things.

src/widgetastic_patternfly/__init__.py
digitronik: For  `spark line charts` we don't have `x-axis` points. Its represent trail of values only....

src/widgetastic_patternfly/__init__.py
digitronik: @mshriver I agreed it will provide `flexibility` and also agreed with `may be a poor page design` possibility. I tried changes locally and  found some observations.

If we add `class` like `chart-pf-sparkline` in `BaseLocator` as  per standard design. most of user need to inherit and create own `BaseLocator`. Above condition they can handle by:
1. We are  providing `locator` as one `argument`. Users can use `locator`; if they found such thing in their page design.
2. If they like to add check  for `class` and don't  want to provide locator for specific graph. they can go with inheritance and overwrite `BASE_LOCATOR` as per their own requirements like....
```
BASE_LOCATOR = ".//div[@id={id} and contains(@class, 'chart-miq-spark')]"
```

What do you think? we should  go with more global approach so user can use widget directly and customize only if they needs.

digitronik:
@mshriver I added changes and questions answers. Please have a look.

src/widgetastic_patternfly/__init__.py
mshriver: Thanks Nikhil - I think that the _default_ behavior of the widget needs to model the patternfly reference implementation, and be as precise as possible in its xpath locators.

For that reason, I would prefer to have the patternfly reference class in the base locator XPATH. Just as you said, if I want to completely override the locator because my project uses a significantly different base element, I can do that directly.  Otherwise, if just the class changes, then subclassing as below is very simple, and explicitly shows how the markup changes for the project where this is used.

```
class MyProjectSparkLine(SparkLineChart):
    BASE_LOCATOR = '//div[@id={} and contains(@class, 'myproject-spark-class')]'
```

PR #79, opened 2018-12-17
[1LP][RFR] read method to Dropdown

digitronik:
-  I observed that for `Dropdown` widget; we need method which will return `currently selected option`.
-  This type  of method is in `SelectedDropdown`. So change name of this method and moved to `Dropdown` class.
- Method name should be same for similar operation among all widgets. 

- We had used this in `SSUIDropdown`[0]. Now,  I need this for `custom button`. For checking the display option of `Group`. Objective is just check name display or not.

[0] https://github.com/ManageIQ/integration_tests/blob/master/widgetastic_manageiq/__init__.py#L1143

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -1538,6 +1538,14 @@ class Dropdown(Widget):
             raise DropdownDisabled('Dropdown "{}" is not enabled'.format(self.text))
 
     @property
+    def currently_selected(self):
+        """Returns the currently selected item text."""
+        return self.browser.text(self.BUTTON_LOCATOR, parent=self)
+
+    def read(self):
+        return self.currently_selected
+
+    @property
     def is_open(self):
         return 'open' in self.browser.classes(self)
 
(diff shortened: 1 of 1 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -1579,6 +1579,11 @@ def has_item(self, item):
         """
         return item in self.items
 
+    @property
+    def selected_option(self):
```
izapolsk: Q: what is returned by read() call for this widget ?

digitronik: No; with read its not working... please give me some time...

digitronik:
@izapolsk added `read` in `Dropdown`  widget and some related changes in `SelectorDropdown`

src/widgetastic_patternfly/__init__.py
```diff
@@ -1537,6 +1537,9 @@ def _verify_enabled(self):
         if not self.is_enabled:
             raise DropdownDisabled('Dropdown "{}" is not enabled'.format(self.text))
 
+    def read(self):
+        return self.browser.text(self.BUTTON_LOCATOR, parent=self)
```
mshriver: I think this is a bit backwards - `read` should call `currently_selected`, which should be defined here in Dropdown.

Both methods should be in Dropdown.

digitronik:
@izapolsk no test available for `Drowpdown` or `SelectDropdown` widget... I will push tests PR; As, I found most of `wt.pfy` widgets have no `unit-test`.

PR #81, opened 2019-01-02
Fix about modal test

digitronik:
1. Added respective `wait_for`
2. `h2` is title tag :)

Optional:
we can go with wait check in `widget` it self but I though better user decide in tests about this.

### DIFF
--- a/testing/test_about_modal.py
+++ b/testing/test_about_modal.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+from wait_for import wait_for
+
 from widgetastic.widget import View
 from widgetastic_patternfly import AboutModal, Button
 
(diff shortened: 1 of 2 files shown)


testing/test_about_modal.py
```diff
@@ -27,15 +31,18 @@ class TestView(View):
...
 
     assert view.about.trademark == trademark
 
-    assert view.about.items()  # TODO: check the dictionary
+    assert view.about.items() == ITEMS
 
     # close the modal
     view.about.close()
+    wait_for(lambda: not view.about.is_open, delay=0.1, num_sec=5)
     assert not view.about.is_open
```
izapolsk: this is not necessary then

testing/test_about_modal.py
```diff
@@ -27,15 +31,18 @@ class TestView(View):
     assert view.button.title == 'Launch Modal'
     assert not view.button.disabled
     view.button.click()
+
     view.flush_widget_cache()
+    wait_for(lambda: view.about.is_open, delay=0.1, num_sec=5)
     assert view.about.is_open
```
izapolsk: this is already checked by wait for

digitronik:
@izapolsk I added changes. If possible please merge this. I would like to replace `twitter-bootstrap` .js  in `DataVisualization` widget PR.

PR #82, opened 2019-01-04
[RFR] Add in a widget for a list-view object

john-dupuy:
Adding a widget for a basic list-view handling class: 

https://www.patternfly.org/pattern-library/content-views/list-view/

This widget is designed to work similarly to `Table` and `TableRow`, except here we have `ItemsList` and `item_class` which is input as a parameter when the list is instantiated. The functionality and filtering is fairly bare-bones at this stage.

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -18,7 +18,6 @@ from widgetastic.exceptions import (
     WidgetOperationFailed,
 )
 from widgetastic.log import call_sig
-from widgetastic.utils import Parameter, ParametrizedLocator, VersionPick, partial_match
 from widgetastic.widget import (
     BaseInput,
     ClickableMixin,
@@ -28,7 +27,9 @@ from widgetastic.widget import (
     View,
     Widget,
     do_not_read_this_widget,
+    ParametrizedView
 )
+from widgetastic.utils import ParametrizedLocator, Parameter, VersionPick, partial_match
 from widgetastic.xpath import quote
 
 from wait_for import wait_for, wait_for_decorator, TimedOutError
@@ -2582,6 +2583,43 @@ class GroupedBarChart(LineChart):
     pass
 
 
+class ListItem(ParametrizedView):
+    """ Basic item object for use with ItemsList"""
+
+    PARAMETERS = ("index",)
+    ROOT = ParametrizedLocator(
+        './/div[contains(@class,"list-group-item") and position()={index}]'
+    )
+    DESCRIPTION_LOCATOR = './/span[contains(@class,"description-column")]'
+    EXPAND_LOCATOR = './/span[contains(@class,"{}")]'.format(PFIcon.icons.ANGLE_RIGHT)
+    COLLAPSE_LOCATOR = './/span[contains(@class,"{}")]'.format(PFIcon.icons.ANGLE_DOWN)
+
+    # note that this has 1-based indexing
+    index = Parameter("index")
+
+    # properties
+    @property
+    def item_list(self):
+        return self.parent
+
+    @property
+    def description(self):
+        desc = self.browser.element(self.DESCRIPTION_LOCATOR, parent=self)
+        return self.browser.text(desc)
+
+    # methods
+    def open(self):
+        expand_arrow = self.browser.element(self.EXPAND_LOCATOR, parent=self)
+        self.browser.click(expand_arrow)
+
+    def close(self):
+        collapse_arrow = self.browser.element(self.COLLAPSE_LOCATOR, parent=self)
+        self.browser.click(collapse_arrow)
+
+    def read(self):
+        return self.browser.text(self)
+
+
 class ItemsList(View):
     """Basic list-view handling class:
     https://www.patternfly.org/pattern-library/content-views/list-view/
@@ -2599,8 +2637,10 @@ class ItemsList(View):
             # define the list-view widget that's on this page
             @View.nested
             class item_list(ItemsList):
-                # define the item_class that the list-view uses
+                # define the item_class that the list-view uses (default is ListItem)
                 item_class = ItemClass
+                # define a default assoc_column that can be used for filtering
+                assoc_column = <assoc_column_str>
             # define whatever else this page has on it ...
 
     For an example: integration_tests/cfme/control/explorer/alerts.py::MonitorAlertsAllView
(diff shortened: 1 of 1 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -2351,3 +2351,111 @@ def item_select(self, item, close=True):
...
+                item_class = ItemClass
+            # define whatever else this page has on it ...
+
+    For an example: integration_tests/cfme/control/explorer/alerts.py::MonitorAlertsAllView
+
+    Usage is as follows assuming the list is instantiated as ``view.item_list``:
+
+    .. code-block:: python
+        # Access item by position
+        view.item_list[0].next() # => gives first item in list-view
```
mshriver: indexed and next call to pull from generator, why is that? I'd expect item_list[0] to give me the first row?

src/widgetastic_patternfly/__init__.py
```diff
@@ -2351,3 +2351,111 @@ def item_select(self, item, close=True):
...
+        # Iterate through rows
+        for item in view.item_list:
+            do_something()
+        # You can also filter items in two main ways
+        # 1) by an assoc_column that corresponds to an attribute of the item_class e.g.
+        item_list = view.item_list
+        item_list.assoc_column = 'description'
```
mshriver: Maybe include this definition in the View definition example above, showing that it can be set when the class is defined, not just when accessing the widget.

src/widgetastic_patternfly/__init__.py
```diff
@@ -2351,3 +2351,111 @@ def item_select(self, item, close=True):
...
+        filtered_items = view.item_list[item_filter]
+        # note that filters can also be applied to the items() method
+        # e.g.
+        filtered_items = view.item_list.items(item_filter)
+
+    Args:
+         assoc_column: Name of an attribute/property defined in the item_class
+    """
+
+    ROOT = './/div[contains(@class,"list-view-pf-view")]'
```
mshriver: ROOT locator should not be relative xpath expression, view instantiation should handle setting the parent browser object so that it is correctly located within the parent View (if this is nested in a main content div view)

src/widgetastic_patternfly/__init__.py
```diff
@@ -2351,3 +2351,111 @@ def item_select(self, item, close=True):
...
+    ITEMS = './/div[contains(@class,"list-group-item-header")]'
+    item_class = None
```
mshriver: We need a basic item_class here I think, so that this is usable out of the box? What would a reference patternfly ItemClass look like?

src/widgetastic_patternfly/__init__.py
```diff
@@ -2351,3 +2351,111 @@ def item_select(self, item, close=True):
...
+
+    def __init__(self, parent, assoc_column=None, logger=None):
+        View.__init__(self, parent, logger=logger)
```
mshriver: Call your parent class constructor via `super()`, not directly.

src/widgetastic_patternfly/__init__.py
```diff
@@ -2351,3 +2351,111 @@ def item_select(self, item, close=True):
...
+        self.assoc_column = assoc_column
+
+    def __getitem__(self, item_filter):
+        """ allows the ability to directly select AlertItem by a filter with
+            item = view.alerts_list[item_filter],
+            item_filter can be of type: string, dict, int, or None
```
mshriver: I think this should explicitly look for these types, and raise ValueError immediately if its not an expected type.

src/widgetastic_patternfly/__init__.py
```diff
@@ -2351,3 +2351,111 @@ def item_select(self, item, close=True):
...
+            value = item_filter.values()[0]
+            if len(item_filter) > 1:
+                self.logger.warning(
+                    "List-view filter currently not implemented for dictionaries"
+                    "greater than a length of one. Selecting the first key value pair from the dict"
+                )
+        # filter via string, note the default
+        elif isinstance(item_filter, str):
+            if self.assoc_column is None:
+                self.assoc_column = "description"  # default behavior
```
mshriver: I'm thinking this behavior should be in a property on the class, so that anywhere that we need to default assoc_column (future methods?)  it will have the same defaulting behavior.

In constructor, set `self._assoc_column`, then have the `assoc_column` property return `self._assoc_column or 'description'`

src/widgetastic_patternfly/__init__.py
```diff
@@ -2351,3 +2351,111 @@ def item_select(self, item, close=True):
...
+    # properties
+    @property
+    def item_count(self):
+        """ returns how many rows are currently in the table."""
+        return len(self.browser.elements(self.ITEMS, parent=self))
+
+    # methods
+    def items(self, item_filter=None):
+        """ returns a generator for all Items matching the item_filter"""
+        start = 1
```
jawatts: why start at 1 here?

src/widgetastic_patternfly/__init__.py
mshriver: I might be wrong here, MRO might be picking up the wrong class.

src/widgetastic_patternfly/__init__.py
john-dupuy: This is because of the fact that the method `items()` is a generator,`__getitem__` is designed to be used with a general `item_filter`, not necessarily just an integer. I agree that if `item_filter` happens to be an integer, we should definitely return that row. This behavior can be included in `__getitem__`

src/widgetastic_patternfly/__init__.py
john-dupuy: Agreed -- this could be as simple as something like the `AlertItem` class without all the specific stuff that is relevant to alerts. Many of the properties will be ItemList specific, but that is okay.

src/widgetastic_patternfly/__init__.py
john-dupuy: @jawatts This is because the `position()` of the items in the list start at 1 and not 0 -- similar to the `Table` and `TableRow`

jawatts: Thanks! I think it is worth to put that is a comment here.

src/widgetastic_patternfly/__init__.py
john-dupuy: Agreed, I will have include this in my updates

src/widgetastic_patternfly/__init__.py
izapolsk: This is correct. I can't recall exact reason but it was necessary in some cases.

PR #88, opened 2019-03-22
[RFR] Add fill method for button

john-dupuy:
Implementing fill method for `Button`.

### DIFF
--- a/testing/test_buttons.py
+++ b/testing/test_buttons.py
@@ -15,3 +15,6 @@ def test_button_click(browser):
     assert view.button1.is_displayed
     assert view.button2.is_displayed
     assert view.button3.is_displayed
+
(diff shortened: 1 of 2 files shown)


mshriver:
Please add to the tests to cover this fill method:
https://github.com/RedHatQE/widgetastic.patternfly/blob/master/testing/test_buttons.py

testing/test_buttons.py
```diff
@@ -15,3 +15,6 @@ class TestView(View):
     assert view.button1.is_displayed
     assert view.button2.is_displayed
     assert view.button3.is_displayed
+
+    FILL_DICT = {"any_button": True, "button1": True, "button2": True, "button3": True}
```
izapolsk: I'd try to pass False at least once :)

john-dupuy: Good idea -- I put it in there!

PR #89, opened 2019-04-30
[RFR] re pattern changes

digitronik:
Signed-off-by: Nikhil Dhandre <ndhandre@redhat.com>

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -1319,13 +1319,13 @@ class BootstrapTreeview(Widget):
                 step = step.pick(self.browser.product_version)
         else:
             image = None
-        if not isinstance(step, six.string_types + (re._pattern_type,)):
+        if not isinstance(step, six.string_types + (Pattern,)):
             step = str(step)
         return image, step
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -36,6 +36,12 @@
 
 from .utils import PFIcon
 
+# py3.7 support
```
mshriver: What is this intending to fix? This newly named `Pattern` isn't used anywhere in the module - and there are a few direct references to `re._pattern_type` here.

digitronik: I missed... let me correct them.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1313,13 +1319,13 @@ def _process_step(self, step):
                 step = step.pick(self.browser.product_version)
         else:
             image = None
-        if not isinstance(step, six.string_types + (re._pattern_type,)):
+        if not isinstance(step, six.string_types + (Pattern,)):
```
digitronik: I'm suspect about this... need deep check. @mshriver your comment on this?

PR #91, opened 2019-05-13
[RFR] hover property for dropdown

digitronik:
Signed-off-by: Nikhil Dhandre <ndhandre@redhat.com>

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -1708,6 +1708,11 @@ class Dropdown(Widget):
                 self.logger.warning('There is an unexpected alert present.')
                 pass
 
+    @property
+    def hover(self):
+        # title will act as hover for disabled Dropdown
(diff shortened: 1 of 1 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -1708,6 +1708,11 @@ def item_select(self, item, handle_alert=None):
                 self.logger.warning('There is an unexpected alert present.')
                 pass
 
+    @property
+    def hover(self):
+        # title will act as hover for disabled Dropdown
+        return self.browser.element(self.BUTTON_LOCATOR).get_attribute("title")
```
john-dupuy: Q: Thoughts on having a `get_hover_text` function that will can take an attribute other than "title"? I have seen some dropdowns have a the title be called something slightly different. This property could then do `self.get_hover_text("title")`.

john-dupuy: Nevermind, got confused between bootstrap select vs dropdown. This LGTM

PR #92, opened 2019-05-16
[RFR]: Add in modal widget with basic functionality

john-dupuy:
Adding a modal widget and a unit test to accompany it: 

https://www.patternfly.org/pattern-library/widgets/#modal

Since the form that is present in the modal can be general, the unit test provides an example use case for dealing with this. The `SpecificModal` inherits the `Modal` class and then overwrites the `body` widget.

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -1935,12 +1935,6 @@ class Modal(View):
         ROOT = './/div[@class="modal-body"]'
         body_text = Text(locator=".//h4")
 
-        def fill(self, value):
-            if not self.form:
-                return False
-            else:
-                return self.form.fill(value)
-
     @View.nested
     class footer(View):  # noqa
         """ The footer of the modal """
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -1897,8 +1897,8 @@ class Modal(View):
             'and contains(@class, "fade") and @role="dialog"]')
 
     def __init__(self, parent, id=None, logger=None):
+        self.id = id
         if id:
-            self.id = id
             self.ROOT = ParametrizedLocator(
                 './/div[normalize-space(@id)={@id|quote} and '
                 'contains(@class, "modal") and contains(@class, "fade") '
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -1893,17 +1893,18 @@ class Modal(View):
 
         https://www.patternfly.org/pattern-library/widgets/#modal
     """
-    ROOT = ParametrizedLocator('.//div[normalize-space(@id)={@id|quote} and '
-                               'contains(@class, "modal") and contains(@class, "fade") '
-                               'and @role="dialog"]')
+    ROOT = ('.//div[contains(@class, "modal") '
+            'and contains(@class, "fade") and @role="dialog"]')
 
     def __init__(self, parent, id=None, logger=None):
+        if id:
+            self.id = id
+            self.ROOT = ParametrizedLocator(
+                './/div[normalize-space(@id)={@id|quote} and '
+                'contains(@class, "modal") and contains(@class, "fade") '
+                'and @role="dialog"]')
+
         View.__init__(self, parent, logger=logger)
-        self.id = id
-        if not self.id:
-            # if no id, we make assumptions about the modal's locator
-            self.ROOT = ('.//div[contains(@class, "modal") '
-                         'and contains(@class, "fade") and @role="dialog"]')
 
     @property
     def title(self):
@@ -1933,8 +1934,6 @@ class Modal(View):
         """ The body of the modal """
         ROOT = './/div[@class="modal-body"]'
         body_text = Text(locator=".//h4")
-        # form widget, this varies from modal to modal, so it is defaulted to None
-        form = None
 
         def fill(self, value):
             if not self.form:
@@ -1956,12 +1955,6 @@ class Modal(View):
     def accept(self):
         """ Submit/Save/Accept/Delete for the modal."""
         self.footer.accept.click()
-
-    def fill(self, value):
-        filled = self.body.fill(value)
-        if filled:
-            self.submit()
-        return filled
 
 
 class BreadCrumb(Widget):
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -1888,6 +1888,82 @@ class AboutModal(Widget):
         return items
 
 
+class Modal(View):
+    """ Patternfly modal widget
+
+        https://www.patternfly.org/pattern-library/widgets/#modal
+    """
+    ROOT = ParametrizedLocator('.//div[normalize-space(@id)={@id|quote} and '
+                               'contains(@class, "modal") and contains(@class, "fade") '
+                               'and @role="dialog"]')
+
+    def __init__(self, parent, id=None, logger=None):
+        View.__init__(self, parent, logger=logger)
+        self.id = id
+        if not self.id:
+            # if no id, we make assumptions about the modal's locator
+            self.ROOT = ('.//div[contains(@class, "modal") '
+                         'and contains(@class, "fade") and @role="dialog"]')
+
+    @property
+    def title(self):
+        return self.header.title.read()
+
+    @property
+    def is_displayed(self):
+        """ Is the modal currently open? """
+        try:
+            return "in" in self.browser.classes(self)
+        except NoSuchElementException:
+            return False
+
+    def close(self):
+        """Close the modal"""
+        self.header.close.click()
+
+    @View.nested
+    class header(View):  # noqa
+        """ The header of the modal """
+        ROOT = './/div[@class="modal-header"]'
+        close = Text(locator='.//button[@class="close"]')
+        title = Text(locator='.//h4[@class="modal-title"]')
+
+    @View.nested
+    class body(View):  # noqa
+        """ The body of the modal """
+        ROOT = './/div[@class="modal-body"]'
+        body_text = Text(locator=".//h4")
+        # form widget, this varies from modal to modal, so it is defaulted to None
+        form = None
+
+        def fill(self, value):
+            if not self.form:
+                return False
+            else:
+                return self.form.fill(value)
+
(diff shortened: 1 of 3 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -1888,6 +1888,62 @@ def items(self):
...
+            # if no id, we make assumptions about the modal's locator
+            self.ROOT = ('.//div[contains(@class, "modal") '
+                         'and contains(@class, "fade") and @role="dialog"]')
+
+    @property
+    def title(self):
+        return self.title_widget.read()
+
+    @property
+    def is_open(self):
```
izapolsk: do you think we need two same properties ?

src/widgetastic_patternfly/__init__.py
```diff
@@ -1888,6 +1888,62 @@ def items(self):
...
 
+class Modal(Widget):
+    """ Patternfly modal widget
+
+        https://www.patternfly.org/pattern-library/widgets/#modal
+    """
+    ROOT = ParametrizedLocator('.//div[normalize-space(@id)={@id|quote} and '
+                               'contains(@class, "modal") and contains(@class, "fade") '
+                               'and @role="dialog"]')
+    CLOSE_LOC = './/div[@class="modal-header"]/button[@class="close"]'
```
izapolsk: I'd create Text widget with `close` name instead.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1888,6 +1888,62 @@ def items(self):
...
+    # widgets for the title, cancel, submit, form widget
+    title_widget = Text(locator='.//h4[@class="modal-title"]')
+    cancel_button = Button(classes="btn-default")
+    submit_button = Button(classes="btn-primary")
+    form_widget = None
```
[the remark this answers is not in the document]
john-dupuy: This is for the general form that may be present in the modal, I will move this into a nested view.

src/widgetastic_patternfly/__init__.py
john-dupuy: Nope, I was copying structure from `AboutModal` but I can get rid of this.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1888,6 +1888,62 @@ def items(self):
         return items
 
 
+class Modal(Widget):
+    """ Patternfly modal widget
+
+        https://www.patternfly.org/pattern-library/widgets/#modal
+    """
+    ROOT = ParametrizedLocator('.//div[normalize-space(@id)={@id|quote} and '
```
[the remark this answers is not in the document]
john-dupuy: After testing this, I think it is better to use the locator I had originally for the following reasons:

1) Having `id` will be the easiest way to distinguish pages that have more than 1 modal, provided that an `id` is present in the markup. 
2) Using the `fade`, `modal` class allows for an easy definition of the `is_displayed` method, as we can simply ask if `in` is in `self.browser.classes(self)`. Using `modal-dialog` we will have to check e.g. the title, which I suppose is fine, but I'd prefer to maintain consistency with the `AboutModal`
3) `AboutModal` and `Modal` should have consistent definitions.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1888,6 +1888,62 @@ def items(self):
...
+
+    @property
+    def is_displayed(self):
+        return self.is_open
+
+    def close(self):
+        """Close the modal"""
+        self.browser.click(self.CLOSE_LOC, parent=self)
+
+    def cancel(self):
```
izapolsk: could you please call these both methods accept and dismiss. it turns out we already use such names in selenium/widgetastic.core.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1888,6 +1888,82 @@ def items(self):
...
+        close = Text(locator='.//button[@class="close"]')
+        title = Text(locator='.//h4[@class="modal-title"]')
+
+    @View.nested
+    class body(View):  # noqa
+        """ The body of the modal """
+        ROOT = './/div[@class="modal-body"]'
+        body_text = Text(locator=".//h4")
+        # form widget, this varies from modal to modal, so it is defaulted to None
+        form = None
```
izapolsk: I'm not sure you need `form` here, body view is form itself.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1888,6 +1888,82 @@ def items(self):
...
+                return False
+            else:
+                return self.form.fill(value)
+
+    @View.nested
+    class footer(View):  # noqa
+        """ The footer of the modal """
+        ROOT = './/div[@class="modal-footer"]'
+        dismiss = Button("Cancel")
+        accept = Button(classes=Button.PRIMARY)
```
izapolsk: could you please explain this Button.PRIMARY property ?

src/widgetastic_patternfly/__init__.py
```diff
@@ -1888,6 +1888,82 @@ def items(self):
...
+
+    def dismiss(self):
+        """ Cancel the modal"""
+        self.footer.dismiss.click()
+
+    def accept(self):
+        """ Submit/Save/Accept/Delete for the modal."""
+        self.footer.accept.click()
+
+    def fill(self, value):
```
izapolsk: I'd leave it remove this method and just let it use default view's fill instead.
It will be more obvious.
 People can easily call self.body.fill directly or pass dict with {'body': {'some_widget': 'some value'}}

testing/test_modal.py
```diff
@@ -0,0 +1,48 @@
...
+    # # open modal again
+    view.button.click()
+    wait_for(lambda: view.modal.is_displayed, delay=0.5, num_sec=5)
+    # make sure buttons are not disabled
+    assert not view.modal.footer.dismiss.disabled
+    assert not view.modal.footer.accept.disabled
+    # make sure the cancel button works
+    view.modal.dismiss()
+
+    wait_for(lambda: not view.modal.is_displayed, delay=0.1, num_sec=5)
```
izapolsk: I guess you need to add at least one case when you accept modal

src/widgetastic_patternfly/__init__.py
```diff
@@ -1888,6 +1888,82 @@ def items(self):
...
+    ROOT = ParametrizedLocator('.//div[normalize-space(@id)={@id|quote} and '
+                               'contains(@class, "modal") and contains(@class, "fade") '
+                               'and @role="dialog"]')
+
+    def __init__(self, parent, id=None, logger=None):
+        View.__init__(self, parent, logger=logger)
+        self.id = id
+        if not self.id:
+            # if no id, we make assumptions about the modal's locator
+            self.ROOT = ('.//div[contains(@class, "modal") '
```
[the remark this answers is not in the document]
john-dupuy: @izapolsk Is it possible use `ParametrizedLocator` after the call to `View.__init__`? Or should I handle this just with a simple `.format`?

src/widgetastic_patternfly/__init__.py
john-dupuy: `Button.PRIMARY` is simply the class `btn-primary`

testing/test_modal.py
john-dupuy: For the reference markup, nothing happens when I click "Save", do we need need to modify the markup to `Save` on click?

izapolsk: it would be great to modify it to fill some input field with any value and check it.

src/widgetastic_patternfly/__init__.py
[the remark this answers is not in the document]
john-dupuy: Nevermind, I think I figured out a good alternative.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1888,6 +1888,69 @@ def items(self):
         return items
 
 
+class Modal(View):
+    """ Patternfly modal widget
+
+        https://www.patternfly.org/pattern-library/widgets/#modal
```
mshriver: :tada:

src/widgetastic_patternfly/__init__.py
```diff
@@ -1888,6 +1888,69 @@ def items(self):
...
+    def is_displayed(self):
+        """ Is the modal currently open? """
+        try:
+            return "in" in self.browser.classes(self)
+        except NoSuchElementException:
+            return False
+
+    def close(self):
+        """Close the modal"""
+        self.header.close.click()
```
mshriver: I think it would be good to wrap this in is_displayed, so that its 'safe' to call when the modal has maybe unexpectedly closed.

@izapolsk thoughts? I can add a commit for this if you agree, since john is on vacation.

izapolsk: I agree. I'm currently fixing one compatibility issue. so, I'll fix this one as well.

src/widgetastic_patternfly/__init__.py
```diff
@@ -1888,6 +1888,74 @@ def items(self):
...
+
+        https://www.patternfly.org/pattern-library/widgets/#modal
+    """
+    ROOT = ('.//div[contains(@class, "modal") '
+            'and contains(@class, "fade") and @role="dialog"]')
+
+    def __init__(self, parent, id=None, logger=None):
+        if id:
+            self.id = id
+            self.ROOT = ParametrizedLocator(
```
izapolsk: I think this also can be combined with default ROOT but this is rather optional.

mshriver: Thought about the same thing - technically could compose either, and we've done so in some other widgets, but it gets pretty ugly. I think in this case what is here is easily replaced by anything inheriting.

I'm more concerned about `self.id` not getting set unless its passed. I'm going to push a commit to this to move setting of `self.id`

src/widgetastic_patternfly/__init__.py
```diff
@@ -1888,6 +1888,74 @@ def items(self):
...
+    def is_displayed(self):
+        """ Is the modal currently open? """
+        try:
+            return "in" in self.browser.classes(self)
+        except NoSuchElementException:
+            return False
+
+    def close(self):
+        """Close the modal"""
+        self.header.close.is_displayed and self.header.close.click()
```
mshriver: :+1: sexy

izapolsk: :)

PR #93, opened 2019-05-24
[WIPTEST] fixing Modal.is_displayed. it shouldn't call ensure_page_safe()

izapolsk:


src/widgetastic_patternfly/__init__.py
```diff
@@ -1919,7 +1919,8 @@ def text(self):
     def is_displayed(self):
         """ Is the modal currently open? """
         try:
-            return "in" in self.browser.classes(self)
```
mshriver: While this is a workaround, I don't think all of this is necessary.

`Browser.get_attribute` is still calling `Browser.element`, which calls `Browser.elements`, resulting in a call to `Browser.plugin.ensure_page_safe` because of the default kwargs.

`Browser.classes()` takes kwargs, and they should be passed all the way down the call chain to `elements`, just like in `get_attribute`.

I think you can achieve the same result by just passing `check_safe=False` to `self.browser.classes()`

izapolsk: this isn't final fix. I just saved my work at some checkpoint in order to continue it on laptop :)
I'm currently torn up between two solutions 
1. quick and easy - make direct selenium's  get_attribute call
2.  long and difficult, passing arguments to all nested methods in wt.core elements.

mshriver: I see how some of the other calls within `elements` would need kwargs too.

izapolsk: the problem is that we cannot pass arguments to some of calls.

PR #94, opened 2019-05-28
[RFR]Updating boostrap nav to include all_options

kedark3:
Adding all options property for BoostrapNav

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -507,10 +507,7 @@ class BootstrapNav(Widget):
     def all_options(self):
         """A property to return the list of options available in the BootstrapNav"""
         b = self.browser
-        return [
-            b.text(el)
-            for el in b.elements(self.ITEM_LOCATOR)
-        ]
+        return [b.text(el) for el in b.elements(self.ITEM_LOCATOR)]
 
     def read(self):
         """Implement read()"""
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -480,6 +480,7 @@ class BootstrapNav(Widget):
     See http://getbootstrap.com/components/#nav for more information on Bootstrap nav components.
     """
     ROOT = ParametrizedLocator('{@locator}')
+    ITEM_LOCATOR = './/li'
     CURRENTLY_SELECTED = './/li[contains(@class, "active")]/a'
     TEXT_MATCHING = './/li/a[text()={txt}]'
     PARTIAL_TEXT = './/li/a[contains(normalize-space(.), {txt})]'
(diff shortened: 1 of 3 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -502,6 +503,15 @@ def currently_selected(self):
         """A property to return the currently selected menu item"""
         return [self.browser.text(el) for el in self.browser.elements(self.CURRENTLY_SELECTED)]
 
+    @property
+    def all_options(self):
+        """A property to return the list of options available in the BootstrapNav"""
+        b = self.browser
+        return [
+            el.text
+            for el in b.elements(self.ITEM_LOCATOR, parent=self)
```
mshriver: Looks good, two things come to mind.

1. Needs unit tests :)  Don't hesitate to ask if you want some guidance on this.

2. Why does `currently_selected` use `self.browser.text(el)` instead of `el.text` - is there a chance that the `text` attribute isn't set on an element?

src/widgetastic_patternfly/__init__.py
```diff
@@ -502,6 +503,15 @@ def currently_selected(self):
         """A property to return the currently selected menu item"""
         return [self.browser.text(el) for el in self.browser.elements(self.CURRENTLY_SELECTED)]
 
+    @property
+    def all_options(self):
+        """A property to return the list of options available in the BootstrapNav"""
+        b = self.browser
```
digitronik: Please avoid catching in a variable if you are not using it multiple time.

src/widgetastic_patternfly/__init__.py
digitronik: @mshriver about `self.browser.text` I think we have a guideline
"instead of `element.text` use `browser.text(element)` (applies for all such circumstances)"

mshriver: @digitronik can you please link to where you got that? I'm also more interested in _why_ that is the guideline.  Preliminary reading suggests it may be differentiating between innerText (what user sees) and textContent (includes non-visible text), but I'm not sure yet.

digitronik: @mshriver https://github.com/RedHatQE/widgetastic.core/blob/master/docs/guidelines.rst
`When using Browser` please clear I'm also confused

kedark3: I will update the text call. I used `parent=self` as it was used https://github.com/RedHatQE/widgetastic.patternfly/blob/47b5e98ec07d0658f6e1a1d9e01812be51833c85/src/widgetastic_patternfly/__init__.py#L1031 

Does it mean that's wrong on L1031 as well?

mshriver: Passing `parent=self` is unnecessary when using `Browser` (in the guidelines Nikhil linked) because it will figure out the parent automatically.

kedark3: Right, I have no objection on that, just curious because there are other places in the file where it is used in browser calls. Can you please check?

src/widgetastic_patternfly/__init__.py
kedark3: this is no longer valid as i am using `b` multiple times now.

digitronik: please manage return in a single line

kedark3: Done.

PR #96, opened 2019-06-13
[1LP][RFR] Add Button.text and Button.read()

### LINKED ISSUE #57: Implement .read() for Button widgets
We should be able to read the text of a Button, if there is text present. If there's no text (rare?), we could return None.

mshriver:
and unit tests!

Fixes #57

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -181,6 +181,15 @@ class Button(Widget, ClickableMixin):
         else:
             return False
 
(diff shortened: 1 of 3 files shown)


mshriver:
@izapolsk I thought the same thing, more or less - that the `self.browser.text(self)` could really be in wt.core. For the moment I decided to proceed with implementing here.

wt.core.Widget read() raises NotImplemented, so changing that to default to returning `text` property would be major change, which needs to be discussed.

PR #98, opened 2019-07-17
[RFR] Update NavDropdown locators

mshriver:
Make the locators actually look at `li` classes, not just presence of a caret for NavDropdown expandable checks.

Use ParametrizedLocator, default the locator to wt.pf reference for NavDropdowns.

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -374,14 +374,15 @@ class FlashMessage(Widget):
 
 class NavDropdown(Widget, ClickableMixin):
     """The dropdowns used eg. in navigation. Usually located in the top navbar."""
-    EXPAND_LOCATOR = './a/span[contains(@class, "caret")]'
+    EXPAND_LOCATOR = ('./a["aria-expanded" and '
+                      '"aria-haspopup" and '
+                      'contains(@class, "dropdown-toggle")]')
 
-    def __init__(self, parent, locator, logger=None):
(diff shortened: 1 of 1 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -374,14 +374,15 @@ def type(self):
...
     """The dropdowns used eg. in navigation. Usually located in the top navbar."""
-    EXPAND_LOCATOR = './a/span[contains(@class, "caret")]'
+    EXPAND_LOCATOR = ('./a["aria-expanded" and '
+                      '"aria-haspopup" and '
+                      'contains(@class, "dropdown-toggle")]')
 
-    def __init__(self, parent, locator, logger=None):
-        Widget.__init__(self, parent, logger=logger)
-        self.locator = locator
+    ROOT = ParametrizedLocator('//nav//li[//a[@id="{@id}"] and contains(@class, "dropdown"")]')
```
valaparthvi: I see an extra double quote.

valaparthvi: Query: I'm guessing now we only need to mention the dropdown ID instead of the whole locator or was that the case before as well?

mshriver: Before it was just taking a locator string, now its defaulting to this locator via ROOT, and the user only has to pass the id for the dropdown.

If their app doesn't have an id for the dropdown, they can subclass and set ROOT as needed - as I've done in integration_tests.

valaparthvi: Okay! I just saw your PR [#9084](https://github.com/ManageIQ/integration_tests/pull/9084) that you raised in integration_tests. Thanks!

PR #99, opened 2019-07-19
[RFR] Drop py2 support

mshriver:
Remove uses of six, update travis/tox config, run 2to3 and remove false-positive list() calls


src/widgetastic_patternfly/__init__.py
```diff
@@ -642,7 +643,7 @@ def nav_item_tree(self, start=None):
             result[item] = sub_items or None
         if result and all(value is None for value in result.values()):
             # If there are no child nodes, then just make it a list
-            result = list(six.iterkeys(result))
+            result = list(result.keys())
```
RonnyPfannschmidt: `.keys()` can be removed

src/widgetastic_patternfly/__init__.py
```diff
@@ -2555,7 +2556,7 @@ def read(self):
         Returns:
             :py:class:`dict` complete data on chart
         """
-        return self._get_data(self._elements.values())
+        return self._get_data(list(self._elements.values()))
```
RonnyPfannschmidt: a iterator is fine here, why list it?

src/widgetastic_patternfly/__init__.py
```diff
@@ -2656,7 +2657,7 @@ def get_data_for_legends(self, *legends):
         """
         self.hide_all_legends()
         self.display_legends(*legends)
-        return self._get_data(self._elements.values())
+        return self._get_data(list(self._elements.values()))
```
RonnyPfannschmidt: dito

src/widgetastic_patternfly/__init__.py
```diff
@@ -2665,7 +2666,7 @@ def read(self):
             :py:class:`dict` complete data on chart
         """
         self.display_all_legends()
-        return self._get_data(self._elements.values())
+        return self._get_data(list(self._elements.values()))
```
RonnyPfannschmidt: dito

src/widgetastic_patternfly/__init__.py
```diff
@@ -2814,8 +2815,8 @@ def items(self, item_filter=None):
         stop = self.item_count + 1
         # filter via key, value pair
         if isinstance(item_filter, dict):
-            key = item_filter.keys()[0]
-            value = item_filter.values()[0]
+            key = list(item_filter.keys())[0]
```
RonnyPfannschmidt: this should use `key, value = next(item_filter.items())`

RonnyPfannschmidt:
@mshriver you might want to take a look at using pyupgrade

we might also want to drop py34

src/widgetastic_patternfly/__init__.py
mshriver: Was just being explicit where the called method might be expecting a list. You're right though, `_get_data` is iterating over it and this is redundant.

src/widgetastic_patternfly/__init__.py
mshriver: nifty, thanks!

mshriver:
@RonnyPfannschmidt because of the use of `re.Pattern` I almost want to jump to py3.7+ _only_, but that might be a bit extreme.

PR #101, opened 2019-07-22
update AboutModal, change locators

mshriver:
title locators changed in CFME, allow h1 or h2 by default
change title and trademark to not use Text widgets
Updated title/trade_mark properties to just find/read the elements directly

Change root locator, make id optional, use it when available.

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -383,7 +383,8 @@ class NavDropdown(Widget, ClickableMixin):
                                '//li[.//a[@id={@id|quote} and contains(@class, "dropdown-toggle")]'
                                ' and contains(@class, "dropdown")]')
 
-    def __init__(self, parent, id, logger=None):
+    def __init__(self, parent, id=None, logger=None):
+        """id is optional to allow for easy subclass change of ROOT"""
         Widget.__init__(self, parent, logger=logger)
         self.id = id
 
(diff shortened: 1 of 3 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -1857,20 +1857,29 @@ class AboutModal(Widget):
...
 
-    def __init__(self, parent, id, logger=None):
+    def __init__(self, parent, id=None, logger=None):
         Widget.__init__(self, parent, logger=logger)
         self.id = id
 
+    def __locator__(self):
+        """If id was passed, parametrize it into a locator, otherwise use ROOT_LOC"""
+        if self.id is not None:
+            return ('//div[normalize-space(@id)="{}" and contains(@class, "modal") and '
```
john-dupuy: Why do we avoid using `.//div[contains(@class, "about-modal-pf")]` if the ID is passed?

src/widgetastic_patternfly/__init__.py
```diff
@@ -1857,20 +1857,29 @@ class AboutModal(Widget):
 
     Provides a close method
     """
-    ROOT = ParametrizedLocator('//div[normalize-space(@id)={@id|quote} and '
-                               'contains(@class, "modal") and contains(@class, "fade")]')
+    # ROOT_LOC only used when id is not passed to constructor
+    ROOT_LOC = ('//div[contains(@class, "modal") and contains(@class, "fade") '
```
john-dupuy: Was this changed to `ROOT_LOC` to match the names of the other constants? Most of the other widgets still use `ROOT`

mshriver: using ROOT is to automagically resolve the `__locator__` method, since I don't intend for that to happen here I'm using ROOT_LOC since I directly implement `__locator__`.

src/widgetastic_patternfly/__init__.py
mshriver: no good reason - I can add this to the ID based locator too.

src/widgetastic_patternfly/__init__.py
john-dupuy: Makes sense, thanks!

PR #106, opened 2019-09-05
[1LP][RFR] Add partial_match in BootstrapNav is_disabled method

valaparthvi:
- Add the ability to pass a partial_match text to BootstrapNav's is_disabled.

Require this change for integration_tests [PR#9266](https://github.com/ManageIQ/integration_tests/pull/9266).

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -562,6 +562,9 @@ class BootstrapNav(Widget):
         """Check if an item is disabled"""
         if text:
             # Check if an item is disabled based on the text of that item
+            if isinstance(text, partial_match):
+                partial_text = text.item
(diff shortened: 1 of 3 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -562,6 +562,9 @@ def is_disabled(self, text=None, **kwargs):
         """Check if an item is disabled"""
         if text:
             # Check if an item is disabled based on the text of that item
+            if isinstance(text, partial_match):
```
mshriver: Looks good!

Please add some assertions (or a new test case) to the unit test for this:

https://github.com/RedHatQE/widgetastic.patternfly/blob/master/testing/test_bootstrap_nav.py#L18

PR #108, opened 2019-09-16
[RFR] Raise WidgetOperationFailed instead of StopIteration in breadcrumb

mshriver:
click_location can raise StopIteration, I think its a lot better to raise WidgetOperationFailed


mshriver:
Not much pressure to push this, I think I'll take some time and write unit tests for breadcrumb and include them here.

PR #109, opened 2019-10-14
[RFR] Add handle_alert kwarg to VerticalNavigation.select()

prichard77:
##  Purpose or Intent

- __Extending__  VerticalNavigation.select() by adding kwarg handle_alert.  This will allow disabling of handle_alert call within select() when called with handle_alert=False.

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -661,14 +661,15 @@ class VerticalNavigation(Widget):
             for el
             in self.browser.elements(self.CURRENTLY_SELECTED, parent=self)]
 
-    def select(self, *levels, **kwargs):
+    def select(self, *levels, handle_alert=True, **kwargs):
         """Select an item in the navigation.
-
         Args:
             *levels: Items to be clicked in the navigation.
-
         Keywords:
-            anyway: Default behaviour is that if you try selecting an already selected item, it will
+            handle_alert(bool): If set to True, will call self.browser.handle_alert to handle
+            alert popups.
+            anyway: Default behaviour is that if you try selecting an already selected item,
+            it will
                 not do it. If you pass ``anyway=True``, it will click it anyway.
         """
         levels = list(levels)
@@ -709,8 +710,9 @@ class VerticalNavigation(Widget):
                 self.logger.debug('finishing the menu selection by clicking on %s', level)
                 # No safety check because previous command did it
                 self.browser.click(link, ignore_ajax=True, check_safe=False)
-                self.browser.handle_alert(wait=2.0, squash=True)
-
+                if handle_alert:
+                    self.browser.handle_alert(wait=2.0, squash=True)
+                    
     def get_child_div_for(self, *levels):
         current = self
         for level in levels:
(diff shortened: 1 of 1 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -661,14 +661,15 @@ def currently_selected(self):
...
-
         Args:
             *levels: Items to be clicked in the navigation.
-
         Keywords:
-            anyway: Default behaviour is that if you try selecting an already selected item, it will
+            handle_alert(bool): If set to True, will call self.browser.handle_alert to handle
+            alert popups.
+            anyway: Default behaviour is that if you try selecting an already selected item,
+            it will
```
john-dupuy: L673 and L672 can fit on the same line I think

src/widgetastic_patternfly/__init__.py
```diff
@@ -661,14 +661,15 @@ def currently_selected(self):
             for el
             in self.browser.elements(self.CURRENTLY_SELECTED, parent=self)]
 
-    def select(self, *levels, **kwargs):
+    def select(self, *levels, handle_alert=True, **kwargs):
```
john-dupuy: I recommend leaving `handle_alert` as a `kwarg` and documenting it like you've done already.

src/widgetastic_patternfly/__init__.py
```diff
@@ -709,7 +710,8 @@ def other_div_displayed():
                 self.logger.debug('finishing the menu selection by clicking on %s', level)
                 # No safety check because previous command did it
                 self.browser.click(link, ignore_ajax=True, check_safe=False)
-                self.browser.handle_alert(wait=2.0, squash=True)
+                if handle_alert:
```
john-dupuy: If `handle_alert` is a `kwarg`

```suggestion
                if kwargs.get('handle_alert', True):
```

src/widgetastic_patternfly/__init__.py
```diff
@@ -661,14 +661,15 @@ def currently_selected(self):
...
-    def select(self, *levels, **kwargs):
+    def select(self, *levels, handle_alert=True, **kwargs):
         """Select an item in the navigation.
-
         Args:
             *levels: Items to be clicked in the navigation.
-
         Keywords:
-            anyway: Default behaviour is that if you try selecting an already selected item, it will
+            handle_alert(bool): If set to True, will call self.browser.handle_alert to handle
```
john-dupuy: If you want to leave this as an argument and not a keyword/kwarg, this should be under L666

src/widgetastic_patternfly/__init__.py
mshriver: I would disagree here, I think its an important kwarg to name. If there's a mismatch of consistency in method declarations and their use of handle_alert has a named kwarg, I think all uses should be updated to be a named kwarg.

src/widgetastic_patternfly/__init__.py
mshriver: +1, let's clean this formatting up and then I'm happy to merge.

src/widgetastic_patternfly/__init__.py
john-dupuy: Then I think `anyway` should be a named kwarg too, otherwise it seems inconsistent.

john-dupuy: Up to @prichard77, I can include a followup PR if needed.

PR #111, opened 2019-11-15
warning property for Input

digitronik:
It required to validate Input.
@izapolsk @mshriver I found this specific to partternfly so I don't think we need it for `TextInput`

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -219,15 +219,26 @@ class Input(TextInput):
 
     Has some additional methods.
     """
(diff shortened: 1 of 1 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -229,6 +229,14 @@ def help_block(self):
         else:
             return self.browser.text(help_block)
 
+    @property
+    def warning(self):
+        try:
+            self.browser.wait_for_element('./following-sibling::div', timeout=3)
```
mshriver: Let's store this locator in `Input.WARNING_LOC` so it can be easily overriden and modified.

PR #112, opened 2019-11-19
[RFR] Add regex capability to match_messages

tpapaioa:
This PR adds the ability to pass a compiled regex as the 'text' parameter for the match_messages, assert_message, and assert_success_message methods of the FlashMessages class.

For example, when testing VM retirement using the 'Time Delay from Now' option, the selected month/week/day/hour time delay values will be added to the date/time at which the page was loaded. The resulting flash message from a successful save is of the form "Retirement date set to 12/19/19 18:31 UTC". But if the time changes from one minute to the next while navigating to the Set Retirement Date page, it is impossible to predict accurately whether the flash message will display a time delay calculated from one or the other minute. Now, with a test like the following, 

```
    flash_regex = re.compile(f"^Retirement date set to (12/19/19 18:31 UTC|12/19/19 18:32 UTC)$")
    view.flash.assert_success_message(flash_regex)
```

a single assertion against a range of dates / flash message texts can be performed.

A couple other small changes:
1.) Fixed several typos in which log messages wrote 'Boostrap tree' instead of 'Bootstrap tree'.
2.) Changed the logging in match_messages somewhat to make it clearer and to indicate when the match is actually an inverse match.
3.) Also added a unit test file, test_flashmessages.py, which tests basic functionality of the FlashMessages methods read, assert_no_error, assert_message, assert_success_message, match_messages, and dismiss.

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -274,25 +274,55 @@ class FlashMessages(Widget):
             message.dismiss()
 
     def match_messages(self, text=None, t=None, partial=False, inverse=False):
+        """
+            Return a list of flash messages matching the specified alert text and type(s)
+
+            Args:
+                text: text to compare against each flash message's text attribute. This parameter
+                 can be either a string, in which case normal string comparison will be performed,
+                 or a compiled regular expression, in which case it will use re.match()
+                 (default: None).
+
+                t: an alert type string or tuple/list/set of types, to compare against each flash
+                 message's type attribute (default: None).
+
+                partial: If partial is False, and if `text` is a string, then a message will be
+                 considered a match if the message's text attribute and `text` are equal. If partial
+                 is True, then `text` needs only be contained in the message's text to be considered
+                 a match. This argument has no effect if `text` is a compiled regular expression
+                 (default: False).
+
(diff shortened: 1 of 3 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -274,25 +274,55 @@ def dismiss(self):
             message.dismiss()
 
     def match_messages(self, text=None, t=None, partial=False, inverse=False):
+        """
+            Return a list of flash messages matching the specified alert text and type(s)
```
mshriver: :+1: :+1: :+1:

src/widgetastic_patternfly/__init__.py
```diff
@@ -2042,7 +2079,7 @@ def click_location(self, name, handle_alert=True):
         try:
             location = next(loc for loc in self._path_elements if br.text(loc) == name)
         except StopIteration:
-            self.logger.exception(f'Given location name [{name}] not found')
```
mshriver: Why remove the use of f-string here? I see you didn't use it above, which is fine, but I don't think its appropriate to remove uses of it without some reason behind it.

mshriver: Ignore me, this was working around the py3.5 support that was still active in master, which has been removed.

PR #114, opened 2019-11-30
Modify BootstrapNav TEXT_DISABLED 

valaparthvi:
The current TEXT_DISABLED doesn't always give the correct xpath, modify it to work with all the cases.

Required for: https://github.com/ManageIQ/integration_tests/pull/9685

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -513,7 +513,7 @@ class BootstrapNav(Widget):
     TEXT_MATCHING = './/li/a[text()={txt}]'
     PARTIAL_TEXT = './/li/a[contains(normalize-space(.), {txt})]'
     ATTR_MATCHING = './/li/a[@{attr}={txt}]'
-    TEXT_DISABLED = './/li[contains(@class, "disabled")]/a[text()={txt}]'
+    TEXT_DISABLED = './/li[contains(@class, "disabled")]/a[contains(text(), {txt})]'
     ATTR_DISABLED = './/li[contains(@class, "disabled")]/a[@{attr}={txt}]'
     VALID_ATTRS = {'href', 'title', 'class', 'id'}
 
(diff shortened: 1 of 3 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -513,7 +513,7 @@ class BootstrapNav(Widget):
     TEXT_MATCHING = './/li/a[text()={txt}]'
     PARTIAL_TEXT = './/li/a[contains(normalize-space(.), {txt})]'
     ATTR_MATCHING = './/li/a[@{attr}={txt}]'
-    TEXT_DISABLED = './/li[contains(@class, "disabled")]/a[text()={txt}]'
+    TEXT_DISABLED = './/li[contains(@class, "disabled")]/a[contains(text(), {txt})]'
```
mshriver: There's an assumption baked into making this a contains match, that I don't think we want.

Note the behavior of the `select` and `is_disabled` methods, and how they inspect the `text` argument to determine if its a partial_match object. If it is, then they use the `PARTIAL_TEXT` locator, otherwise the `TEXT_MATCHING` locator.

I think you need to apply the same behavior here, and define a separate `TEXT_PARTIAL_DISABLED` locator, used when the passed text argument is partial match.

src/widgetastic_patternfly/__init__.py
```diff
@@ -514,6 +514,7 @@ class BootstrapNav(Widget):
     PARTIAL_TEXT = './/li/a[contains(normalize-space(.), {txt})]'
     ATTR_MATCHING = './/li/a[@{attr}={txt}]'
     TEXT_DISABLED = './/li[contains(@class, "disabled")]/a[text()={txt}]'
+    PARTIAL_TEXT_DISABLED = './/li[contains(@class, "disabled")]/a[contains(text(), {txt})]'
```
john-dupuy: I don't think it will make any difference, but for consistency with `PARTIAL_TEXT` can we do? 

```suggestion
    PARTIAL_TEXT_DISABLED = './/li[contains(@class, "disabled")]/a[contains(normalize-space(.), {txt})]'
```

mshriver: Good point @john-dupuy, this consistency in normalizing the space should be kept.

valaparthvi: I think I tried with `normalize-space(.)` at first, but it didn't work as expected. Let me check again though.

valaparthvi: So I tested again, and it seems to be working. Pushed the changes.

valaparthvi:
Hey @mshriver can we merge this now?

PR #116, opened 2019-12-11
[RFR] Use XPath instead of Selenium WebElements for FlashMessage locator

tpapaioa:
The FlashMessage class currently uses Selenium WebElements to implement its widget locator:

```
class FlashMessages(Widget):
    @property
    def messages(self):
        result = []
        msg_xpath = ('.//div[@id="flash_text_div" or '
                     'contains(@class, "flash_text_div")]/div[contains(@class, "alert")]')
        try:

            for flash_div in self.browser.elements(msg_xpath, parent=self, check_visibility=True):
                result.append(FlashMessage(self, flash_div, logger=self.logger))

class FlashMessage(Widget):
    def __init__(self, parent, flash_div, logger=None):
        Widget.__init__(self, parent, logger=logger)
        self.flash_div = flash_div

    def __locator__(self):
        return self.flash_div
```

This results in warnings logged to cfme.log any time a FlashMessage is read, e.g.,

```
2019-12-11 10:24:20,477 [I] [cfme] [RequestsView/flash]: Performing exact match of flash messages, text: 'Retirement initiated for 1 VM and Instance from the CFME Database', type: 'success' (.cfme_venv/lib64/python3.7/site-packages/widgetastic_patternfly/__init__.py:315)
2019-12-11 10:24:20,620 [W] [cfme] [RequestsView/flash]: __locator__ of FlashMessage class returns a WebElement! (.cfme_venv/lib64/python3.7/site-packages/widgetastic/widget/base.py:341)
```

This PR re-implements FlashMessages as a View containing FlashMessage as a ParametrizedView, to use XPath locators instead. Each FlashMessage is indexed by the XPath index of the message within the containing FlashMessages block.

Because the user can dismiss these messages in random order, this XPath index is not necessarily fixed. For example, if there are three messages on the page,

text: "Message A", locator: (msg_xpath)[1]
text: "Message B", locator: (msg_xpath)[2]
text: "Message C", locator: (msg_xpath)[3]

and the user dismisses the one with text "Message A", then the remaining messages will now have new locators:

text: "Message B", locator: (msg_xpath)[1]
text: "Message C", locator: (msg_xpath)[2]

To account for this, FlashMessages.messages yields the individual FlashMessage instances one at a time, and the XPath index of the next message is re-calculated in case of message dismissal.

Views that use FlashMessages will need to be updated since it is no longer a Widget:

```
- flash = FlashMessages('.//div[@id="flash_msg_div"]')
+ flash = View.nested(FlashMessages)
```

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -241,187 +241,6 @@ class Input(TextInput):
             return None
 
 
-class FlashMessages(Widget):
-    """Represents the block of flash messages."""
-    def __init__(self, parent, locator, logger=None):
-        Widget.__init__(self, parent, logger=logger)
-        self.locator = locator
-
-    def __locator__(self):
-        return self.locator
-
-    def read(self):
-        result = []
-        for message in self.messages:
-            result.append(message.read())
-        return result
-
-    @property
-    def messages(self):
-        result = []
-        msg_xpath = ('.//div[@id="flash_text_div" or '
-                     'contains(@class, "flash_text_div")]/div[contains(@class, "alert")]')
-        try:
-
-            for flash_div in self.browser.elements(msg_xpath, parent=self, check_visibility=True):
-                result.append(FlashMessage(self, flash_div, logger=self.logger))
-        except NoSuchElementException:
-            pass
-        return result
-
-    def dismiss(self):
-        for message in self.messages:
-            message.dismiss()
-
-    def match_messages(self, text=None, t=None, partial=False, inverse=False):
-        """
-            Return a list of flash messages matching the specified alert text and type(s)
-
-            Args:
-                text: text to compare against each flash message's text attribute. This parameter
-                 can be either a string, in which case normal string comparison will be performed,
-                 or a compiled regular expression, in which case it will use re.match()
-                 (default: None).
-
-                t: an alert type string or tuple/list/set of types, to compare against each flash
-                 message's type attribute (default: None).
-
-                partial: If partial is False, and if `text` is a string, then a message will be
-                 considered a match if the message's text attribute and `text` are equal. If partial
-                 is True, then `text` needs only be contained in the message's text to be considered
-                 a match. This argument has no effect if `text` is a compiled regular expression
-                 (default: False).
-
-                inverse: If inverse is False, then the match is performed as described above. If
-                 inverse is True, then the matching logic is inverted, and only messages that fail
-                 to match will be returned.
-
-            Returns:
-                List of matching FlashMessage objects.
-        """
-        msg_type = t if isinstance(t, (tuple, list, set, type(None))) else (t, )
-        # If inverse is True, further comparison statements will be treated as is.
-        # Otherwise, they will be inverted by not_().
-        op = bool if inverse else not_
-        log_inverse = "inverse " if inverse else ""
-        log_type = f", type: {t!r}" if msg_type else ""
-        if isinstance(text, Pattern):
-            log_part = "pattern"
-            log_text = f", pattern: {text.pattern!r}"
-        else:
-            log_part = "partial" if partial else "exact"
-            log_text = f", text: {text!r}" if text else ""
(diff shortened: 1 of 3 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -246,10 +246,17 @@ class FlashMessages(Widget):
     def __init__(self, parent, locator, logger=None):
         Widget.__init__(self, parent, logger=logger)
         self.locator = locator
+        self.MSG_XPATH = ('.//div[@id="flash_text_div" or '
```
mshriver: I don't think this needs to be defined in the constructor - it can just be a class attribute.

Design wise - that allows for easy modification by sub-classes for non-reference PF implementations.

src/widgetastic_patternfly/__init__.py
```diff
@@ -246,10 +246,17 @@ class FlashMessages(Widget):
...
+                          'contains(@class, "flash_text_div")]/div[contains(@class, "alert")]')
+        self._flash_messages = []
```
mshriver: Up for implementing this as a property with a setter?

src/widgetastic_patternfly/__init__.py
mshriver: Second thought - This is really a property of the individual messages.

`FlashMessages` class could have a `message_widget_class` attribute, and we can reference `self.message_widget_class.LOCATOR` where needed.

That way the locator for the individual message is defined with the individual message class, and its still straightforward to subclass and replace the locator, or the widget class.

src/widgetastic_patternfly/__init__.py
mshriver: Actually, why do we need to store this at all?  I think its going to cause us problems with dismissal and stale references.

src/widgetastic_patternfly/__init__.py
tpapaioa: Re-done to use ParametrizedView for each FlashMessage, indexed by its order in the FlashMessages block (updated PR description).

src/widgetastic_patternfly/__init__.py
tpapaioa: Fixed.

src/widgetastic_patternfly/__init__.py
```diff
@@ -2919,3 +2738,211 @@ def items(self, item_filter=None):
...
+    """
+    TYPE_MAPPING = {
+        'alert-warning': 'warning',
+        'alert-success': 'success',
+        'alert-danger': 'error',
+        'alert-info': 'info'
+    }
+
+    PARAMETERS = ('index',)
+    ROOT = ParametrizedLocator('.//div[contains(@class, "alert") and position()={index}]')
```
mshriver: :ok_hand:

src/widgetastic_patternfly/__init__.py
```diff
@@ -2919,3 +2738,211 @@ def items(self, item_filter=None):
...
+
+    @property
+    def msg_count(self):
+        return len(self.browser.elements(self.MSG_LOCATOR, parent=self))
+
+    def messages(self, **msg_filter):
+        """Return a generator for all notifications matching the msg_filter.
+        The total number of notifications is re-checked each time, and the parametrized XPath
+        index is re-calculated if necessary, in case any of the previously-yielded
+        notifications have been dismissed.
```
mshriver: :+1:

testing/testing_page.html
```diff
@@ -798,35 +798,27 @@ <h4 class="modal-title" id="myModalLabel">Modal Title</h4>
         <span class="pficon pficon-ok"></span>
         <strong>Retirement date set to 12/31/19 15:55 UTC</strong>
       </div>
-    </div>
-    <div class="flash_text_div">
```
mshriver: Does this change match the reference implementation of PF3, or something from MIQ?

mshriver: Got an answer offline, its a bit of both, but inline with reference code from Patternfly docs.

PR #119, opened 2020-06-03
[WIP] - fixing locator of flash message

dgaikwad:
Fixed locator related error:
Was getting below error in testcase:
```
                       f"Expected: {e_text}.\n "
                       f"Available messages: {self.read()}")
>           raise AssertionError(log_msg)
E           AssertionError: Assert exact match of message.
E            Expected: Error: import processing failed: domain: *.
E            Available messages: []

/var/ci/cfme_venv3/lib64/python3.7/site-packages/widgetastic_patternfly/__init__.py:2945: AssertionError
```

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -2800,7 +2800,10 @@ class FlashMessage(ParametrizedView):
 class FlashMessages(View):
     """Represent the div block containing the individual inline notifications."""
     ROOT = './/div[@id="flash_msg_div"]'
-    MSG_LOCATOR = './div[contains(@class, "flash_text_div")]/div[contains(@class, "alert")]'
+    MSG_LOCATOR = (
+        '//div[contains(@class, "flash_text_div") or contains(@class, "flash-message")]'
+        '/div[contains(@class, "alert")]'
+    )
     msg_class = FlashMessage
 
     def __getitem__(self, msg_filter):


src/widgetastic_patternfly/__init__.py
```diff
@@ -2800,7 +2800,8 @@ def type(self):
 class FlashMessages(View):
     """Represent the div block containing the individual inline notifications."""
     ROOT = './/div[@id="flash_msg_div"]'
-    MSG_LOCATOR = './div[contains(@class, "flash_text_div")]/div[contains(@class, "alert")]'
+    MSG_LOCATOR = (f'//div[contains(@class, "flash_text_div") or '
```
digitronik, line 2803: Q: why you need `f-sting`?

dgaikwad: mistakenly added, now removed.

src/widgetastic_patternfly/__init__.py
```diff
@@ -2800,7 +2800,8 @@ def type(self):
 class FlashMessages(View):
     """Represent the div block containing the individual inline notifications."""
     ROOT = './/div[@id="flash_msg_div"]'
-    MSG_LOCATOR = './div[contains(@class, "flash_text_div")]/div[contains(@class, "alert")]'
+    MSG_LOCATOR = ('//div[contains(@class, "flash_text_div") or '
+                   'contains(@class, "flash-message")]/div[contains(@class, "alert")]')
```
valaparthvi, line 2804:
```suggestion
    MSG_LOCATOR = (
        '//div[contains(@class, "flash_text_div") or contains(@class, "flash-message")]'
        '/div[contains(@class, "alert")]'
    )
```

dgaikwad: done

src/widgetastic_patternfly/__init__.py
```diff
@@ -2800,7 +2800,10 @@ def type(self):
 class FlashMessages(View):
     """Represent the div block containing the individual inline notifications."""
     ROOT = './/div[@id="flash_msg_div"]'
-    MSG_LOCATOR = './div[contains(@class, "flash_text_div")]/div[contains(@class, "alert")]'
+    MSG_LOCATOR = (
+        '//div[contains(@class, "flash_text_div") or contains(@class, "flash-message")]'
```
mshriver, line 2804: The xpath shouldn't be changed to be a non-relative locator. The syntax was `./div[...]` because its relative to the ROOT locator, and the methods within the class using this locator are doing it relative to `parent=self`.

Just adding ` or contains(@class, 'flash-message')` to the specification for the first div should be sufficient to match.

tpapaioa: What is the test case that is failing? Does this need to be a change to FlashMessages itself, or if it's an issue only with certain pages/views, could you instead have a subclass of FlashMessages with a custom MSG_LOCATOR for use in the specific views that have this problem?

digitronik: @tpapaioa issue is with some test cases but its more general solution. As per my personal observation @mshriver suggestion is right and it should not effect on current test cases/implementation. `ROOT` set your working area for specific widget (with browser instance) i.e. `flash_text_div` so class attribute `flash-message` web element should work without any trouble.

tpapaioa: Which tests/views have this issue? Why was `div[contains(@class, "alert")]` removed and the relative `'./div'` changed to `'//div'` at the start of `MSG_LOCATOR`?

mshriver: @tpapaioa double check your diff view, `div[contains(@class, "alert")]` is still part of the locator

mshriver: Also note this case is a _little_ screwy in the first place, because flash message is an MIQ thing, Patternfly3 just has the alert divs for `Inline Notifications` widget.

So 'FlashMessage' in wt.pf should really be `InlineNotification`, and widgetastic_manageiq module within integration_tests should define `FlashMessages` parent.  The message locator here is really relevant to widgetastic_manageiq, as this isn't a Patternfly class.

That being said I don't want this PR to turn into removing FlashMessages from the library, which is why I'm open to accepting more classes for the div xpath for individual messages. Anything other than the alert class isn't strictly Patternfly3 anyway.

tpapaioa: I see the `alert` class in the diff now. I still think the relative locator should be kept at the beginning of the locator, and it would be good to know what views have this issue. I don't see 'flash-message' in any of the manageiq ui repositories....

valaparthvi: > Which tests/views have this issue?

@tpapaioa this test - [test_upload_blank_file[datastore]](https://github.com/ManageIQ/integration_tests/blob/48464ecdc292e119a87bde8678cd8363ce32782d/cfme/tests/automate/test_file_import.py#L55). There might be other tests as well, but I'm not aware of them.

tpapaioa: That test is failing to read the alert messages in the `AutomateImportExportView` class in `cfme/automate/import_export.py`, which already overrides the default `FlashMessages` class with the `V2VFlashMessages` subclass, and implements its own `ROOT` attribute:

```
class AutomateImportExportView(AutomateImportExportBaseView):
    @View.nested
    class flash(V2VFlashMessages):
        ROOT = 'div.import-flash-message'
```
It would make more sense to override `MSG_LOCATOR` here as well. This patch fixes the test in my manual testing:

```
diff --git a/cfme/automate/import_export.py b/cfme/automate/import_export.py
index 16cb23a13..618d8de05 100644
--- a/cfme/automate/import_export.py
+++ b/cfme/automate/import_export.py
@@ -42,6 +42,7 @@ class AutomateImportExportView(AutomateImportExportBaseView):
     @View.nested
     class flash(V2VFlashMessages):
         ROOT = 'div.import-flash-message'
+        MSG_LOCATOR = './div[contains(@class, "alert")]'
 
     @View.nested
     class import_file(View):  # noqa
```

The non-relative xpath locator `//` in your patch is effectively overriding `ROOT` and including it as part of your `MSG_LOCATOR`. They should be kept separate.

digitronik: @tpapaioa nice debug thanks. so its page specifc. I would say @dgaikwad close this PR and add one custom widget in `widgetastic_manageiq`.

I would not use `V2VFlashmessages` widget as it was solution for `TEXT_LOCATOR` change. As i can see `strong` tag in locator better just overwrite `Flashmessages` class ROOT and MSG_LOCATOR.

```
class ImportExportFlashMessages(FlashMessages):
    ROOT = ".//div[contains(@class, 'import-flash-message')]"
    MSG_LOCATOR = './/div[contains(@class, "alert")]'
```

dgaikwad:
Thank you @digitronik @tpapaioa @mshriver @valaparthvi for your great review. After all discussion, I have raised new PR(https://github.com/ManageIQ/integration_tests/pull/10173) in `integration_test` to fix this issue and closing this one.

PR #120, opened 2020-06-15
[RFR] Fix problem with locator in SettingsNavDropdown

jaryn:
This requires https://github.com/RedHatQE/widgetastic.patternfly/pull/122 for tests to fully finish passing.

This pr mitigates the problem with `repr` discovered in CFME `integration_tests` and adding tests for such problems for all the widgets in `widgetastic.patternfly`.

Note this PR adds a test of `stringification` of (mostly) all the components defined in the `widgetastic_patternfly`. Perhaps it would be good to come-up with a way to put the tooling for this to Widgetastic itself and let all the descendant projects benefit from it somehow.

Fixes: https://github.com/ManageIQ/integration_tests/issues/10191

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -345,7 +345,7 @@ class NavDropdown(Widget, ClickableMixin):
         self.browser.click('./ul/li[normalize-space(.)={}]'.format(quote(item)), parent=self)
 
     def __repr__(self):
-        return '{}({!r})'.format(type(self).__name__, self.locator)
+        return '{}(id={!r})'.format(type(self).__name__, self.id)
 
 
 class BootstrapNav(Widget):
--- a/testing/test_all_components.py
+++ b/testing/test_all_components.py
@@ -1,0 +1,112 @@
+"""
+This module contains code for auto-discovering Widget classes, instantiating
+them in a programatically defined view class as a new type, which is then
+instantiated and then testing behavior of the widget objects that appear
+to get instantiated upon access.
+"""
+
+from widgetastic.widget import View, Widget
+import widgetastic_patternfly as wp
+import inspect
+import pytest
+import re
+
+
+collected_widgets = {
+    name: t for name, t in inspect.getmembers(wp)
+    if (inspect.isclass(t) and issubclass(t, Widget)
+        and not issubclass(t, wp.ParametrizedView))}
+""" Subclasses of Widget from the widgetastic_patternfly module (including
+it's own imports), that are gonna be tested. Not including the
+ParametrizedView. """
+
+
+DUMMY_NAME = "name_of_the_dummy"
+DUMMY_ID = "id_of_the_dummy"
+DUMMY_LOCATOR = "/dummy"
+DUMMY_TREE_ID = "SOME_DUMMY_TREE_ID"
+
+
+def camel_to_snake(s):
+    """ Converts camel-cased string to snake form (undersored lowercase)
+    Taken from https://stackoverflow.com/a/1176023/1950100 """
+    return re.sub(r'(?<!^)(?=[A-Z])', '_', s).lower()
+
+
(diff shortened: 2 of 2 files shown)


src/widgetastic_patternfly/__init__.py
```diff
@@ -254,7 +255,7 @@ class NavDropdown(Widget, ClickableMixin):
 
     def __init__(self, parent, id=None, logger=None):
         """id is optional to allow for easy subclass change of ROOT"""
-        Widget.__init__(self, parent, logger=logger)
+        GenericLocatorWidget.__init__(self, parent, locator=None, logger=logger)
```
tpapaioa, line 258: I think a better fix would be just to modify the `__repr__` method for this class, so that it uses the `id` attribute, which is the attribute used to instantiate `NavDropdown` and `SettingsNavDropdown` instances. The following patch works in my testing:

```
$ git diff
diff --git a/src/widgetastic_patternfly/__init__.py b/src/widgetastic_patternfly/__init__.py
index b8432fe..ebd6eea 100644
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -345,7 +345,7 @@ class NavDropdown(Widget, ClickableMixin):
         self.browser.click('./ul/li[normalize-space(.)={}]'.format(quote(item)), parent=self)
 
     def __repr__(self):
-        return '{}({!r})'.format(type(self).__name__, self.locator)
+        return '{}(id={!r})'.format(type(self).__name__, self.id)
 
 
 class BootstrapNav(Widget):
```

Before the patch:

```
-> view = navigate_to(appliance.server, 'LoggedIn')
(Pdb) n
[...]
(Pdb) view.help
*** AttributeError: 'NavDropdown' object has no attribute 'locator'

(Pdb) view.settings
*** AttributeError: 'SettingsNavDropdown' object has no attribute 'locator'
```

After:

```
-> view = navigate_to(appliance.server, 'LoggedIn')
(Pdb) n

(Pdb) view.settings
SettingsNavDropdown(id='dropdownMenu2')

(Pdb) view.help
NavDropdown(id='help-menu')
```

jaryn: I think you are right. I did it as you suggested.

testing/test_all_components.py
```diff
@@ -0,0 +1,112 @@
...
+ParametrizedView. """
+
+
+DUMMY_NAME = "name_of_the_dummy"
+DUMMY_ID = "id_of_the_dummy"
+DUMMY_LOCATOR = "/dummy"
+DUMMY_TREE_ID = "SOME_DUMMY_TREE_ID"
+
+
+def camel_to_snake(s):
```
mshriver, line 30: https://github.com/RedHatQE/widgetastic.core/blob/master/src/widgetastic/utils.py#L493

Consider using this `wt.core` util instead, if it will fit the use case here.

jaryn: The camel_to_snake dealt well with the spaced strings:
```
In [6]: re.sub(r'(?<!^)(?=[A-Z])', '_', "CamelToSSSNake").lower()                             
Out[6]: 'camel_to_s_s_s_nake'
```

But this test is not about attribute names, so I changed this as you suggested.

testing/test_all_components.py
```diff
@@ -0,0 +1,107 @@
+"""
+This module contains code for auto-discovering Widget classes, instantiating
+them in a programatically defined view class as a new type, which is then
+instantiated and then testing behavior of the widget objects that appear
+to get instantiated upon access.
+"""
+
+from widgetastic import attributize_string
```
mshriver, line 8: I believe this import has to come from `widgetastic.utils` - this is failing travis right now.

Though I'm not yet sure why travis results aren't being reported on your PR.

mshriver:
Travis unit tests passing, but no check is getting reported to GitHub.

PR #127, opened 2020-11-05
adding separate ROOT to resolve #125 issue

omkarkhatavkar:
Fixed #125

### Test Result 
```
testing/test_date_picker.py .                                                          [100%]

====================================== warnings summary ======================================
testing/test_data_visualization.py:51
  /home/okhatavk/Learning/widgetastic.patternfly/testing/test_data_visualization.py:51: PytestCollectionWarning: cannot collect test class 'TestDataVisualization' because it has a __init__ constructor (from: testing/test_data_visualization.py)
    class TestDataVisualization(View):

-- Docs: https://docs.pytest.org/en/stable/warnings.html
=================== 1 passed, 99 deselected, 1 warning in 76.34s (0:01:16) ===================

```

PR #128, opened 2020-11-05
fixed the modal test, resolved #126 Issue  

### LINKED ISSUE #126: Modal in unit test does not pass is_displayed
```
______________________________ test_generic_modal ______________________________
browser = <testing.conftest.CustomBrowser object at 0x7f5c3316c7f0>
    def test_generic_modal(browser):
        class TestView(View):
            """ Dummy page matching testing_page.html elements"""
            button = Button('Launch demo modal')
            modal = SpecificModal(id=modal_id)

        view = TestView(browser)
        assert not view.modal.is_displayed
        # Open the modal
        assert not view.button.disabled
        view.button.click()
        wait_for(lambda: view.modal.is_displayed,

>                delay=0.5, num_sec=MODAL_INTERACTION_TIMEOUT)
E       wait_for.TimedOutError: Could not do 'lambda defined as `wait_for(lambda: view.modal.is_displayed,
E                    delay=0.5, num_sec=MODAL_INTERACTION_TIMEOUT)`' at /home/travis/build/RedHatQE/widgetastic.patternfly/testing/test_modal.py:50 in time
```

The id is passed for the modal, and the `if id` block is evaluated in the modal constructor, but the ROOT locator does not appear to be correctly updated.

```
(Pdb) view.modal.__locator__()
Locator(by='xpath', locator='.//div[contains(@class, "modal") and contains(@class, "fade") and @role="dialog"]')
```

The modal, though displayed, is not reading the `'in`' class that is included in the modal `div` when the modal is opened.

```
(Pdb) view.modal.browser.classes(view.modal)
{'modal', 'fade'}
```

omkarkhatavkar:
Fixed https://github.com/RedHatQE/widgetastic.patternfly/issues/126

### Test Result 
```
========================================================= test session starts ==========================================================
platform linux -- Python 3.7.3, pytest-6.1.2, py-1.9.0, pluggy-0.13.1
cachedir: .tox/py37/.pytest_cache
rootdir: /home/okhatavk/Learning/widgetastic.patternfly
plugins: httpserver-0.3.6, cov-2.10.1, allure-pytest-2.8.19
collected 100 items / 99 deselected / 1 selected                                                                                       

testing/test_modal.py 
-- Docs: https://docs.pytest.org/en/stable/warnings.html
======================================== 1 passed, 99 deselected, 1 warning in 92.26s (0:01:32) ========================================
```
_

### DIFF
--- a/src/widgetastic_patternfly/__init__.py
+++ b/src/widgetastic_patternfly/__init__.py
@@ -1808,13 +1808,16 @@ class Modal(View):
 
     def __init__(self, parent, id=None, logger=None):
         self.id = id
-        if id:
-            self.ROOT = ParametrizedLocator(
-                './/div[normalize-space(@id)={@id|quote} and '
-                'contains(@class, "modal") and contains(@class, "fade") '
-                'and @role="dialog"]')
-
         View.__init__(self, parent, logger=logger)
+
+    def __locator__(self):
+        """If id was passed, parametrize it into a locator, otherwise use ROOT"""
+        if self.id is not None:
+            return ('//div[normalize-space(@id)="{}" and contains(@class, "modal")'
+                    'and contains(@class, "fade") and @role="dialog"]'
+                    .format(self.id))
+        else:
+            return self.ROOT
 
     @property
     def title(self):


omkarkhatavkar:
Travis Issue fixed in PR #127

src/widgetastic_patternfly/__init__.py
```diff
@@ -1808,14 +1808,17 @@ class Modal(View):
...
         self.id = id
-        if id:
-            self.ROOT = ParametrizedLocator(
-                './/div[normalize-space(@id)={@id|quote} and '
-                'contains(@class, "modal") and contains(@class, "fade") '
-                'and @role="dialog"]')
-
         View.__init__(self, parent, logger=logger)
 
+    def __locator__(self):
```
digitronik, line 1813: In `View` or `Widget` when we set `ROOT` locator should auto-resolve.

I think it will be great if we fallow an approach like 
```
ROOT = ParametrizedLocator("{@locator}")
def __init__(self, parent, id=None, logger=None):
    self.id = id
    if id:
        self.locator = <some_loator_with id>
    else:
        self.locator = <default>
```

src/widgetastic_patternfly/__init__.py
```diff
@@ -1808,14 +1808,17 @@ class Modal(View):
...
+        """If id was passed, parametrize it into a locator, otherwise use ROOT"""
+        if self.id is not None:
+            return ('//div[normalize-space(@id)="{}" and contains(@class, "modal")'
```
mshriver, line 1816: Please continue to use ParametrizedLocator here, its handling the injection of `self.id` into the string. No need to use `format` here, that part shouldn't change.

src/widgetastic_patternfly/__init__.py
mshriver: I think implementing `__locator__` here is okay, though the pattern @digitronik suggested above does it without implementing the dunder method.

Why its broken in the first place, and why this fix works, is because of the WidgetMetaclass behavior in `__new__`.

Before the WidgetDescriptor instance is created when the unit test is defining the test view class, WidgetMetaclass is setting `__locator__` based on the presence of `ROOT` in the kwargs.  After this, when the WidgetDescriptor resolves into a View instance in the presence of a browser, `self.ROOT` is modified - but `self.__locator__` has already been created automatically based off of the original definition.

mshriver (changes requested):
See specific comments please, sorry for _massive_ delay on review.

mshriver:
@omkarkhatavkar keep an eye here: https://github.com/RedHatQE/widgetastic.patternfly/pull/132/files  

I added a skip for this test so that I can see all passing, you can remove the skip in your branch if that gets merged before you make an update here. Otherwise I'll remove the skip marker.

PR #132, opened 2022-02-14
Move to GitHub Workflow

mshriver:
Modernize a bit, bringing things up-to-date with wt.core standards.

Use pre-commit.  wt.core is using mypy, which caused some failures, so I excluded it (for now).

Remove tox and travis and use a github workflow for the PR and deploy patterns.

## Testing

I opened this branch as a PR against my own fork, where the workflow will actually execute. Should all be green there now, giving me confidence the pull request workflow will work here.

The deploy workflow has a secret that needs to be added to GH for pypi push.

https://github.com/mshriver/widgetastic.patternfly/pull/1

## Docs

I've cleaned up and updated the sphinx conf, but the indicies aren't building correctly. Its the same behavior as what's in wt.pf4, and I'd like to handle fixing both of them in a PR other than this one.


.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,35 @@
...
+        run: |
+          python -m pip install --upgrade setuptools wheel twine
+          python setup.py sdist bdist_wheel
+          python -m twine check dist/*
+
+      - name: Deploy to PyPi
+        uses: pypa/gh-action-pypi-publish@v1.5
+        with:
+          user: __token__
+          password: ${{ secrets.pypi_wt_pf }}
```
mshriver, line 35: Secret added at repository level

PR #136, opened 2022-05-10
Update for selenium4 endpoint, lock image tag

mshriver:


testing/test_data_visualization.py
```diff
@@ -46,7 +46,7 @@
 BAR_LEGENDS = ["Q1", "Q3", "Q2", "Q4"]
 
 
-class TestDataVisualization(View):
+class DataVisualization(View):
```
mshriver, line 49: This class starting with `Test` makes pytest try to read and instantiate it like a test class.

PR #137, opened 2022-05-10
Remove aenum dependency

mshriver:
Using standard python enums requires referencing name/value attribute
directly, which aenum.Constant made a bit easier.

I'd rather not have the dependency, and update callers to use
PFIcon.icons new enums.

Downstream dependency wise, airgun isn't using PFIcon directly, and neither is insights-qe.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -18,8 +18,7 @@ classifiers=
 [options]
 python_requires = >= 3.8
 install_requires =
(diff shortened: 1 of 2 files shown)


JacobCallahan:
Can't really ACK on this repo, but the change looks sane to me

quarckster (approved):

PR #139, opened 2022-05-10
Update Description-Content-Type, set maintainer

mshriver:


setup.cfg
```diff
@@ -2,9 +2,11 @@
 name = widgetastic.patternfly
 author = Milan Falesnik
 author_email = mfalesni@redhat.com
+maintainer = RedHatQE
+maintainer_email = mshriver@redhat.com
 description= Patternfly widget library for Widgetastic
 long_description = file: README.rst
-long_description_content_type='text/x-rst'
+long_description_content_type=text/x-rst
```
mshriver, line 9: The type itself is correct, workflow failures indicate the single quotes were getting included in the value.

